PR #3, opened 2015-02-12
[WIP] Transforms

kaedroho:
This pull request adds a new API called "Transforms"

``Transform`` objects descibe the scaling, offsetting, rotation and sizing to perform on an image. It also introduces an operation called ``transform`` that applies a ``Transform`` to an image.

Advantages:
 - Only runs one operation on the image, no matter how much cropping, resizing, rotating, etc is going on
 - ``Transform`` objects can be used for transforming points as well (in both directions). This is useful for finding where a focal point lies in a particular rendition for example (which would be handy for frontend image cropping).
 - Backends no longer need to implement ``get_size`` as this is now tracked by the ``Transform``. This also means that Wagtails image operations can be run without touching an image file
 - We now have an object to test against in Wagtails image operations (not that there's anything wrong with the way we test image operations at the moment)

Example:

```python
transform = Transform(640, 480).crop(0, 0, 100, 100).offset(0, 50).resize(50, 50)
image.transform(transform) # Transform the image

x, y = transform.atob(100, 100) # Transform coordinates from original image into rendition
x, y = transform.btoa(100, 100) # Vice versa
```

TODO:
 - [ ] Wand support
 - [ ] Implement ``atob`` and ``btoa`` methods (bonus)
 - [ ] Rotation and flipping (and possibly include a solution for #2) (bonus)
 - [ ] Implement some resizing methods on ``Image`` that use transforms (bonus)

### DIFF
--- a/willow/backends/pillow.py
+++ b/willow/backends/pillow.py
@@ -60,6 +60,21 @@ def pillow_crop(backend, left, top, right, bottom):
     backend.image = backend.image.crop((left, top, right, bottom))
 
 
+@PillowBackend.register_operation('transform')
+def pillow_transform(backend, transform):
+    Image = backend.get_pillow_image()
+
+    # We transform 4 times too big then downsize with Image.resize as that has a
+    # much better downsampling filter
+    transform = transform.resize(transform._width * 4, transform._height * 4)
+    m = transform._matrix
+    backend.image = backend.image.transform(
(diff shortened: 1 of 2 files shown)


kaedroho:
Closing this for now. Transforms will be implemented as an internal optimisation after #13 is merged (and will be used automatically).

PR #8, opened 2015-03-21
Allow cascade to be specified for detect_faces

### LINKED ISSUE #5: Need ability to try or specify alternate face detection
We have lots of standing full height (head to feet) photos where the faces are detected better with "haarscascade_frontface_alt_tree.xml". Currently, "haarscascade_frontalface_alt2.xml" is hardcoded and results in the focal point rect being larger than the image in many cases.

For what its worth I'm not sure if this need is shared by others but since we have over 150,000 images I imagine having some configuration options and/or ability to specify fall backs would be good. 

Does this make sense in any way?

kaedroho:
Fixes #5

### DIFF
--- a/willow/backends/opencv.py
+++ b/willow/backends/opencv.py
@@ -62,14 +62,17 @@ def detect_features(backend):
 
 
 @OpenCVBackend.register_operation('detect_faces')
-def detect_faces(backend):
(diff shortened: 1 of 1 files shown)


zerolab:
Out of curiosity, does this need a test?

kaedroho:
This would be a very tricky one to test with unit tests...

PR #13, opened 2015-05-11
Backends refactor

### LINKED ISSUE #1: Make buffers a bit smarter
Currently, buffers only store a single frame in RGB format.

This can cause the following problems:
 - Animation cannot be transferred between backends
 - Transparency gets lost
 - Images not in RGB format may lose colour data

kaedroho:
This pull request makes some major changes to the way Willow works internally. It should make the internals much easier to understand and also pave way for some new features/optimisations/fixes.

**What we have now**

Currently, Willow has a few backend classes which are registered against the ``Image`` class. The ``Image`` class is responsible for discovering the operations in each backend and deciding which backend to use and when to switch.

 - When an image is loaded, it must always be in a backend. So it is always loaded by either Pillow or Wand regardless of what the user wants to do with it (which can be poor for performance if it makes the wrong decision)
 - To switch backends, ``Image`` tells the current backend to put the images pixel data into a buffer and the new backend to load it (which has limitations as descibed in #1)
 - The syntax for creating backends and registering operations against them isn't very nice

**What this refactor changes**

Most of the old code has been binned and replaced with a new model based on three things: states, converters and operations.

Each state class represents a particular way an image could be represented in Willow. There are state classes for each image library the image could be loaded into, and also state class for file formats and intermediate buffers.

An instance of a state class is an actual image. So an instance of ``PillowImageState`` for example, is an image loaded into pillow.

When an image is first opened, one of the ``*ImageFileState`` classes are used to represent the image. (eg, ``.jpg`` files will initially be loaded into a ``JPEGImageFileState`` instance)

Here's a list of states that are provided out of the box:
```
# willow.states.files
# File states used to represent images on the disk before loading them
JPEGImageFileState
PNGImageFileState
GIFImageFileState

# willow.states.buffers
# Buffer states used for converting between two states that are not aware of each other
RGBImageBufferState
RGBAImageBufferState

# willow.plugins.pillow
PillowImageState

# willow.plugins.wand
WandImageState

# willow.plugins.opencv
OpenCVColorImageState
OpenCVGrayscaleImageState
```

States are linked together with ``converters``. These are functions that take an instance of one state class, converts the image and returns an instance of the new state class. For example, a converter from ``JPEGImageFileState`` to ``PillowImageState`` would be a function that calls ``PIL.Image.open`` to open the file and create a Pillow image, this image then gets put into a new instance of ``PillowImageState`` and returned.

Operations are linked to the states. For example, you can register a ``resize`` operation against ``PillowImageState`` which would call Pillows ``Image.resize`` operation internally.

All of these things are registered in the Willow registry (which can be found at ``willow.registry``). This allows external plugins to add more state classes, file formats, operations, etc. The registry is also responsible for planning routes between states.

Converters and operations are usually methods on the state class but they can also be registered separately.

The ``Image`` class is still provided (without any public API changes) but its implemention is greatly simplified by the new registry.

Some advantages of the new way:
 - Images are not immediately loaded into an image library allowing Willow to decide which image library to use based on the operations the user requests (eg, images can now be loaded into OpenCV directly instead of being forced to go through Pillow first). This improves performance
 - Specialised converters can be made between backends. Improving performance and removing the risk of data being lost by the buffers (fixes #1)
 - Backends can have as many states as they like. This allows OpenCV to have both colour and greyscale representations. It also allows us to create states for lazily running operations which could massively improve performance.

TODO:
 - [x] Deactivate states when backend library is not present
 - [x] Better planning (think of a way for it to decide the best state to go to. allow it to make more than one hop)
 - [x] Apply https://github.com/torchbox/Willow/pull/23
 - [x] Reimplement ``original_format`` property
 - [ ] Tests for Image class


kaedroho:
This just needs some routing code (for planning paths from one state to another) and tests and it will be ready to merge.

PR #14, opened 2015-05-21
Image orientation

kaedroho:
This pull request addresses #2.

This is built on my earlier backends refactor branch so the relevant commits are right at the end of the below log.


kaedroho:
I'd like to refactor this into its own operation before merging

PR #15, opened 2015-06-02
Transparency info for a gif is a tuple, pillow expects an integer

maraujop:
A Gif real `image.info` dictionary looks like:

```
{
    'version': 'GIF89a',
    'extension': ('XMP DataXMP', 814),
    'transparency': (151, 239, 255),
    'duration': 30,
    'background': 255,
    'loop': 0
}
```

However pillow expects transparency to be an integer, not a tuple of three integers. This number doesn't seem to affect the generated output, so I'm dropping transparency handling, which might not be the right thing to do.

You can easily reproduce this uploading a gif89a to wagtail under images and then listing images.

Thanks, cheers
Miguel

### DIFF
--- a/willow/backends/pillow.py
+++ b/willow/backends/pillow.py
@@ -77,10 +77,7 @@ def save_as_png(backend, f):
 
 @PillowBackend.register_operation('save_as_gif')
 def save_as_gif(backend, f):
-    if 'transparency' in backend.image.info:
-        backend.image.save(f, 'GIF', transparency=backend.image.info['transparency'])
-    else:
-        backend.image.save(f, 'GIF')
+    backend.image.save(f, 'GIF')
 
 
 @PillowBackend.register_operation('has_alpha')


kaedroho:
Thanks @maraujop.

I think it was implemented that way so gifs keep their transparency when saved (which Pillow doesn't seem to do automatically).

Is it possible that you could provide example images that are misbehaving? Also, what version of Pillow are you using?

maraujop:
Thanks for your fast response @kaedroho ,

I'm attaching you a gif that is misbehaving:
1mg1vfm

Pillow version is 2.8.1 which I believe is the latest stable version. 

I'm not sure how to switch over the wand backend or if wagtail has a settings for configuring which image backend is installed. I've got wand installed and doesn't seem to be using that backend.

Thanks, cheers
Miguel

shredding:
I stumbled across the problem and here's a fix: https://github.com/torchbox/Willow/pull/23

shredding:
@maraujop : Mind checking if https://github.com/torchbox/Willow/commit/c6746a5bb496bb9c94f6314e55c7e58c63f45d47 fixed your issue? I think it's the same bug.

kaedroho:
Seems to work on latest master so looks like #23 did fix it. Thanks everyone!

PR #17, opened 2015-08-13
Try checking image format as a stream

sixpearls:
Sometimes, the file that Willow will get is a BytesIO stream rather than a path string. Since `imghdr` can determine the image format from the stream as well, I figure we might as well try it.


kaedroho:
Thanks @sixpearls 

Could you add a test and a comment in the code as well?

sixpearls:
Hi @kaedroho sure, I just pushed them up. Let me know what you think.

zerolab:
#96 switched to using [filetype](https://pypi.org/project/filetype/) and lots of tests now use BytesIO. Will close this PR on that basis.

Issue #18, opened 2015-08-25
Alpha gets lost when resizing image


nwjlyons:
Hi Torchbox,

If the image is mode `P` ([8-bit pixels, mapped to any other mode using a colour palette](http://pillow.readthedocs.org/en/latest/handbook/concepts.html#modes)) the image gets [converted](https://github.com/torchbox/Willow/blob/master/willow/backends/pillow.py#L52) to `RGB` which drops the alpha channel.

With the alpha channel lost any transparency appears as solid black.

It works if the image is converted to `RGBA` but I'm just wondering does the image need to be converted at all?

shredding:
This is leading to a serious bug with gif's because RGB changes the transparency to a tuple, e.g. `(291,0,0)`.

This will break here: https://github.com/python-pillow/Pillow/blob/master/PIL/GifImagePlugin.py#L426

`int() argument must be a string or a number, not 'tuple'`

zerolab:
Fixed by #23

PR #19, opened 2015-08-25
Retain alpha channel when resizing image.

nwjlyons:
If the image is mode P ([8-bit pixels, mapped to any other mode using a colour palette](http://pillow.readthedocs.org/en/latest/handbook/concepts.html#modes))
the image gets converted to RGB which drops the alpha channel.

With the alpha channel lost any transparency appears as solid black.

In this pull request I'm not converting the image. Let me know if this is a bad idea.

#18

### DIFF
--- a/willow/backends/pillow.py
+++ b/willow/backends/pillow.py
@@ -50,7 +50,7 @@ def get_size(backend):
 
 @PillowBackend.register_operation('resize')
 def resize(backend, size):
-    if backend.image.mode in ['1', 'P']:
+    if backend.image.mode in ['1']:
         backend.image = backend.image.convert('RGB')
 
     backend.image = backend.image.resize(


kaedroho:
Thanks @nwjlyons 

This conversion (originally introduced in: https://github.com/torchbox/verdant-rca/commit/877ef72ffd7c94c57a6c39155b48f91bab44bab3#diff-a824596ff511f2f7019cbb7a7009b899R8) was added to work around Pillow's poor filtering of ``P`` images when resizing down.

Maybe we should convert ``P`` images to ``RGBA`` instead?  We could use the ``has_alpha`` check to decide whether to convert it to ``RGB`` or ``RGBA`` (https://github.com/torchbox/Willow/blob/master/willow/plugins/pillow.py#L33).

kaedroho:
Superseded by #38

PR #21, opened 2015-09-26
Let PIL save the icc profil

bloodywing:
Some images have icc profiles embed. PIL would just strip them without icc_profile as parameter.


bloodywing:
No longer valid. See #63 for a better implementation.

Issue #10, opened 2015-04-01
Add possibility for image optimisation


zerolab:
Could piggyback off http://jpegclub.org/jpegtran/, http://pmt.sourceforge.net/pngcrush/ or similar libraries

balazs-endresz:
Doesn't mod_pagespeed do all that already?

zerolab:
It does indeed.

Well, I suppose it makes sense to defer it to other tools. @kaedroho?

kaedroho:
Does mod_pagespeed run on the fly as the image is served?

balazs-endresz:
> ngx_pagespeed speeds up your site and reduces page load time by automatically applying web performance best practices to pages and associated assets (CSS, JavaScript, images) _without requiring you to modify your existing content or workflow_

But I've also heard that it does this asynchronously, so I'm not sure if the first visitor gets the optimised version too.

davecranwell:
I don't feel mod_pagespeed etc is the correct place for optimisation in this instance. Mod_pagespeed is a multitool for people who don't necessarily have the knowledge/time to fix issues directly.

According to Google Page Speed literally all the images uploaded by users to any given Wagtail site are wasteful (in terms of lossless optimisations possible) when served to visitors. If _all_ images are wasteful I feel we need a more deliberate approach. Given Wagtail's growing reputation for being easy and intuitive, I believe we should be addressing optimisation proactively ourselves.

gasman:
Looks like this is what [mozjpeg](https://github.com/mozilla/mozjpeg) is designed for. There's some interest in adding support for it to Pillow, but no sign of it happening yet: https://github.com/python-pillow/Pillow/issues/539

davecranwell:
The tool I'd normally run by hand to optimise static assets uses a combination of http://pmt.sourceforge.net/pngcrush/ (as @zerolab mentioned), http://optipng.sourceforge.net/, http://www.kokkonen.net/tjko/projects.html and http://www.lcdf.org/gifsicle/

davecranwell:
Just a note that JPG and PNG optimisation should be the primary focus. JPG particularly as it's used for all photographs, which I'd guesstimate account for 98% of the images on our sites.

alexfromvl:
How about making a module with automatic overstore in **webp**. Here is an excellent example https://github.com/andrefarzat/django-webp of a controller based on Django. I think it can be applied to wagtail cms and improve the overall performance of the image.

PR #26, opened 2015-11-23
[WIP] Image optimisation

kaedroho:
Addresses #10

I plan on adding support for at least one optimiser for png and jpeg. The implemention will be done in a way that we can support multiple optimisers and choose one based on installation status/performance. It'll also provide extra operations to allow calling a specific optimiser directly (allowing the developer to pass in any custom arguments).

### DIFF
--- a/willow/image.py
+++ b/willow/image.py
@@ -10,6 +10,9 @@ class UnrecognisedImageFormatError(IOError):
 
 
 class Image(object):
+    # See comment in ImageFile class for description of this attribute
+    allow_automatic_conversion = True
+
     @classmethod
     def check(cls):
         pass
@@ -139,6 +142,13 @@ class RGBAImageBuffer(ImageBuffer):
 
 class ImageFile(Image):
     format_name = None
+
+    # It never makes sense to automatically convert to a file class
+    # This prevents Willow from converting images to files when a file-specific
+    # operation (eg, optimise) is mistakenly performed on the image.
+    # In this case, a user must explicitly convert to a file class first by
+    # using a save_as_* operation first. Eg: "i.save_as_png().optimise()"
+    allow_automatic_conversion = False
 
     def __init__(self, f):
         self.f = f
--- a/willow/image.py
+++ b/willow/image.py
@@ -153,6 +153,11 @@ class ImageFile(Image):
     def __init__(self, f):
         self.f = f
 
+    @Image.operation
+    def optimise(self):
+        # By default, do nothing
+        return self
+
 
 class JPEGImageFile(ImageFile):
     format_name = 'jpeg'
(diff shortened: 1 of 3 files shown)


bashu:
+1

kaedroho:
After some testing, it doesn't look like optipng does much better than Pilllow itself.

One example PNG image I have is **410178 bytes** long (It's the first image to come up in a Google search for "Bender Rodriguez").
I ran optipng on this with it's default settings. This cut the image down to **98542 bytes**
But when I tried just loading and saving that image with Pillow (no operations performed), it is cut down to **94755 bytes**.

So in this case, Pillow on its own performs better than optipng!

elalienx:
Hi! How about Huffman tables lossless jpeg optimisation that would reduce more kb because are used more on each webpage.
By the way excellent work on creating a wrapper to handle images on Wagtail!

kaedroho:
Just been doing some investigation into JPEG image optimisation. Here are my results.

I tested jpegoptim and jpegtran on the JPEG images used on the torchbox.com website. I ran the following commands on each file:

``jpegoptim --strip-all --strip-iptc --strip-icc -d{out} {in}``
``jpegtran -optimize -progressive -outfile {out} {in}``

The output results were exactly the same, they both reduced 196M of images (2103 files) down to 129M of progressive JPEGs, which is a 35% average reduction!

ddiazpinto:
It is not necessary to use another library to optimize images. PIL can do that pretty well. The problem is other. I have been testing the performance of this wrapper and the differences using PIL directly and I found some big differences in the results.

For example. I've tested a JPEG image of 1900x720 (424,5 KB) using by default 85 of quality. Using `fill-1920x400` tag in wagtail throws two operations: `FillOperation()` and `WidthHeightOperation`. The result is an image of 1900x386 and 282,6 KB. Using the same API function of PIL directly (`convert('RGB')`, `crop(Rect(0, 162, 1900, 558)`, `resize((1900, 396), Image.ANTIALIAS)` and `save(name, quality=85)`) the result is an image of 1900x396 (obviously) but 156,4 KB (~45% of improvement).

Using other options of `save()` function like `optimized=True` and `quality=False`, I get an image of 115,6 KB. This is about 60% of improvement.

Obviously, there is no appreciable differences between any of this images so, the conclusion is that something wrong it happens in an step before PIL, presumably with this library.

Any idea of what?

Anyway, I think that is necessary to allow this library to use all the options of each function of PIL, but this is a transversal matter that implies also Wagtail. How do you see it?

elalienx:
Hi @ddiazpinto do you test if [Google Page Speed Insights](https://developers.google.com/speed/pagespeed/insights/) counts the PIL optimised images as compliant according to their test?
Thanks in advance.

kaedroho:
That sounds very promising, I'll look into it when I next get time to work on this, thanks @ddiazpinto!

kaedroho:
Implemented in #39

PR #28, opened 2015-12-09
Add an automatic image orientation function and its test units

serzans:


willow/plugins/pillow.py
```diff
@@ -94,6 +94,31 @@ def open(cls, image_file):
         image_file.f.seek(0)
         image = _PIL_Image().open(image_file.f)
         image.load()
+
+        # JPEG files can be orientated using an EXIF tag.
+        # Make sure this orientation is applied to the data
+        if isinstance(image_file, JPEGImageFile):
```
kaedroho: This code should be moved into a new ``auto_orient`` operation

kaedroho:
Merged in f8a890f3f08ab36ad606d64d82a26d844d3a4834 + parents. Thanks!

PR #29, opened 2015-12-18
Documentation

kaedroho:
First pass at the documentation. Not aiming for perfection yet.

- [x] Sphinx setup
- [x] Intro
- [x] Concepts
- [x] Installation
- [x] General usage
- [x] Extensions
- [x] Reference


docs/installation.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Willow supports Python 2.7+ and 3.3+. It's a pure-python library with no hard
+dependencies so doesn't require a C compiler for a basic installation.
+
+Installation using ``pip``
+--------------------------
+
+.. code-block:: shell
+
+    pip install Willow==0.3b3
```
zerolab: Quick note, this should now be "Willow==0.3". Could [|release|](http://www.sphinx-doc.org/en/stable/config.html#confval-release) do the trick here?

kaedroho: Probably not since the release is currently ``0.4a0``. Probably best to remove the version altogether now

PR #33, opened 2016-01-21
support opencv 3.

wgiddens:
This pull request adds support for OpenCV 3. It will fall back to the old cv api if cv2 is not available.


wgiddens:
I also fixed a random typo in registry.py

willow/plugins/opencv/__init__.py
```diff
@@ -0,0 +1,11 @@
+try:
```
kaedroho: Could the modules be renamed to ``opencv2`` and ``opencv3``?

willow/plugins/opencv/opencv1.py
```diff
@@ -1,57 +1,34 @@
...
-    def has_alpha(self):
-        # Alpha is not supported by OpenCV
-        return False
-
-    @Image.operation
-    def has_animation(self):
-        # Animation is not supported by OpenCV
-        return False
-
+        cv
```
kaedroho: I don't think there's a need to reference ``cv`` here. ``pass`` should do.

willow/plugins/opencv/base.py
```diff
@@ -0,0 +1,65 @@
...
+class BaseOpenCVColorImage(BaseOpenCVImage):
+    @classmethod
+    def from_buffer_rgb(cls, image_buffer):
+        raise NotImplemented()
+
+    # TODO: Converter back to RGBImageBuffer
+
+
+class BaseOpenCVGrayscaleImage(BaseOpenCVImage):
+    min_size = (40, 40)
```
kaedroho: These attribute names don't make much sence outside of the ``detect_faces`` method. Could you move them back or change their names to make their use more clear? (eg ``min_face_size``)

willow/plugins/opencv/base.py
```diff
@@ -0,0 +1,65 @@
...
+    haar_scale = 1.1
+    min_neighbors = 3
+    haar_flags = 0
+
+    def detect_features(self):
+        raise NotImplemented()
```
kaedroho: This should be ``raise NotImplementedError``

kaedroho:
Thanks @wgiddens, This looks pretty good!

Thanks for spotting that typo, I've merged that onto master already (1c7a2948e4073e42fedc1b5ea3eb9234794b0915).

As for the OpenCV 3 support, I've added some inline comments above. Please look into those if you have time. I'll look into getting OpenCV 3 running on Drone.

wgiddens:
I believe I've made the changes you mentioned.  If I missed one let me know.

kaedroho:
Both OpenCV 2 and 3 are importable on ``cv`` and ``cv2`` so the OpenCV 3 backend is actually being used for both versions.

But the OpenCV 3 code you've written is also compatible with OpenCV 2. So I think we should recombine the two plugins into one using your OpenCV 3 implementation. Does that sound OK?

wgiddens:
Fine by me.  My goal is to use wagtail/willow with python 3 which requires opencv 3 and the cv2 interface.  If you are fine with dropping cv, I am.

I probably won't be able to look at it for a few more days while I get caught back up at work.

wgiddens:
Alright, that should do it.

I also made a small edit to the docker file.

nadersoliman:
One question, will OpenCV 2.4.x work with Python3.5 at least for the time being. I have tried to build it but so far it doesn't seem to work (libraries import issues)

kaedroho:
@nadersoliman OpenCV 2.x doesn't support Python 3 yet. But the code in this PR may work though (but there's likely to be bugs).

wgiddens:
I noticed 0.4 is out.  Are there any plans on merging this in or supporting Python 3 some other way?

kaedroho:
> I noticed 0.4 is out. Are there any plans on merging this in or supporting Python 3 some other way?

I hope to get this merged, just need to find a bit of time to review/test the PR.

Will hopefully be able to have a look at this next week.

kaedroho:
Merged in https://github.com/wagtail/Willow/commit/e720c213605a5c62bc08d66fc414820dcc81f739

Thanks!

PR #36, opened 2016-03-22
Validate EXIF data before transpose

findsje:
I have some images returning zero from exif orientation data


kaedroho:
Thanks for the fix @findsje. I've merged this into both ``master`` and ``stable/0.3.x``

PR #44, opened 2016-09-26
Fix spelling of 'optimize' kwarg

gasman:
As pointed out at https://github.com/torchbox/Willow/pull/39#r61911040, but never got fixed...


kaedroho:
Thanks!

PR #52, opened 2017-04-06
Add rotate operation for Willow - Images

mrchrisadams:
Hi there,

I'm opening a PR here, partly to start some work being able to work on issue #2901 on the main wagtail repo.

#https://github.com/wagtail/wagtail/issues/2901

As I understand it at the mo, I'd need to implement rotate()  transformations along the lines of:

```python
@Image.operation
    def rotate(self, size):
        image = self.image
        return PillowImage(image.rotate(90, expand=True))
```

…for Pillow, and Wand, before I can call these higher up the stack from within Wagtail, in the image editing view, 

Would you let me know if this is the case? If so, I'll implement the other rotate method, and start on making the necessary changes for a PR on Wagtail too.

### DIFF
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -192,6 +192,17 @@ class PillowImage(Image):
 
         return cls(image)
 
+    @Image.operation
+    def rotate(self, angle):
+        image = self.image
+
+        if angle > 360:
+            angle = 360
+        elif angle < -360:
+            angle = -360
+        
+        return PillowImage(self.image.rotate(angle))
+
     @Image.converter_to(RGBImageBuffer)
     def to_buffer_rgb(self):
         image = self.image
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -12,7 +12,8 @@ from willow.image import (
     RGBAImageBuffer,
 )
 
-
+class UnsupportedRotation(Exception): pass
+    
 def _PIL_Image():
     import PIL.Image
     return PIL.Image
@@ -62,6 +63,33 @@ class PillowImage(Image):
     @Image.operation
     def crop(self, rect):
         return PillowImage(self.image.crop(rect))
+
+    @Image.operation
+    def rotate(self, angle=90):
+        """
+        Accept a multiple of 90 to pass to the underlying Pillow function
+        to rotate the image.
+        """
+
+        Image = _PIL_Image()
+        ORIENTATION_TO_TRANSPOSE = {
+            90: Image.ROTATE_90,
+            180: Image.ROTATE_180,
+            270: Image.ROTATE_270
+        }
+        transpose_code = ORIENTATION_TO_TRANSPOSE.get(angle)
+
+        if not transpose_code:
+            raise UnsupportedRotation(
+                "Sorry - we only support rotations by 90, 180, or 270 degrees"
+            )
+
+        # We call "transpose", as it rotates the image,
+        # updating the height and width, whereas using 'rotate'
+        # only changes the contents of the image.
+        rotated = self.image.transpose(transpose_code)
+
+        return PillowImage(rotated)
 
     @Image.operation
     def set_background_color_rgb(self, color):
@@ -191,17 +219,6 @@ class PillowImage(Image):
         image.load()
 
         return cls(image)
-
-    @Image.operation
-    def rotate(self, angle):
-        image = self.image
-
-        if angle > 360:
-            angle = 360
-        elif angle < -360:
-            angle = -360
-        
-        return PillowImage(self.image.rotate(angle))
 
     @Image.converter_to(RGBImageBuffer)
     def to_buffer_rgb(self):
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -75,13 +75,20 @@ class PillowImage(Image):
         ORIENTATION_TO_TRANSPOSE = {
             90: Image.ROTATE_90,
             180: Image.ROTATE_180,
-            270: Image.ROTATE_270
+            270: Image.ROTATE_270,
         }
-        transpose_code = ORIENTATION_TO_TRANSPOSE.get(angle)
+
+        modulo_angle = angle % 360
+
+        # is we're rotating a multiple of 360, it's the same as a no-op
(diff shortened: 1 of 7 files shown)


kaedroho:
Hi @mrchrisadams. Yep, you just need to implement the ``rotate`` method on each image class that supports rotation.

I think we should allow the rotation and expand parameters to be specified, and these would also need to be implemented for the Wand version.

Thanks for working on this!

mrchrisadams:
Hi @kaedroho, I know it's been _almost two years_ since I touched this PR, but I've finally found the time to implement the image rotation methods to allow for server-side rotation, in 90 degree increments.

### Why just 90 degree increments

I've just chosen the 90 degree increments as this is the precursor to some work I'm doing with @simo97 on implementing a image rotation in the Wagtail  UI itself, and we're aiming to just support the most common use cases for image rotation to begin with.

It turns out that Wand and Pillow handle image rotation differently, which is why I've kept the API simple - just accepting a multiple of 90 degrees for a rotation.

As far as I can tell, [PIL uses a `transpose` to rotate an image from landscape to portrait format, and rotate to move the contents of the image](https://pythontic.com/image-processing/pillow/rotate).

This is different to wand, which allows for arbitrary rotations, but takes an [optional background channel colour param when you have non-right angle rotations](http://docs.wand-py.org/en/0.5.1/guide/transform.html#rotation), in case you want to fill in the colour), and then resizes the box accordingly.

### Conventions and compatibility

Anyway, please let me know if there's anything else you need - I've tried to to follow the conventions I saw in the repo, as this is my first PR, I may well have missed something obvious.

Also, I might need some pointers how rotation might work with the auto-orientation methods - at present, this doesn't touch exif data.

Ta,

mrchrisadams:
Actually, I've just looked over the history and I realise it's a it messy - I'll rebase locally and push a few tidier commits.

mrchrisadams:
Okay' that should be a tider PR now, without the messy rebase artifacts.

Can you let me know if I should add to the changelog, or is that something the maintainer does when a new release is published?

tomdyson:
Thanks @mrchrisadams! I've requested a review from @kaedroho.

willow/plugins/pillow.py
```diff
@@ -63,6 +64,33 @@ def resize(self, size):
     def crop(self, rect):
         return PillowImage(self.image.crop(rect))
 
+    @Image.operation
+    def rotate(self, angle=90):
```
kaedroho: I think we should remove the default value for the angle.

willow/plugins/pillow.py
```diff
@@ -63,6 +64,33 @@ def resize(self, size):
...
+        to rotate the image.
+        """
+
+        Image = _PIL_Image()
+        ORIENTATION_TO_TRANSPOSE = {
+            90: Image.ROTATE_90,
+            180: Image.ROTATE_180,
+            270: Image.ROTATE_270
+        }
+        transpose_code = ORIENTATION_TO_TRANSPOSE.get(angle)
```
kaedroho: Might be nice to add ``% 360`` to the angle here

mrchrisadams: Ah.

I left 360 out, as if you rotate an image 360 degrees, you rotate it _all_ the way around, and it ends up looking exactly the same as before.

That's my thinking at least - @simo97 's originally had a check if the angle was greater than 360, or less than -360, and you're right it might be worth covering the other angles someone might conceivably add.

Is there a use case you had in mind here, where you can imagine deliberately rotating an image 360 degrees?

willow/plugins/pillow.py
mrchrisadams: No worries - will update this later today.

willow/plugins/pillow.py
kaedroho: I actually meant that we should perform a modulo 360 operation on the angle, so that an input of 450 would be translated to 90. This would also have the effect of converting 360, 720 and -360 to 0 so these can all be handled generically.

> Is there a use case you had in mind here, where you can imagine deliberately rotating an image 360 degrees?

My thinking is that it's possible someone would add two angles together (or hit the rotate 90 button more than 3 times) and the developer forget to add ``% 360`` into their code themselves.

kaedroho:
I have thought about this a bit before and have some ideas for optimisations we can make that might benefit your image editing UI.

1) We can perform EXIF orientation and user orientation in a single pass

Maybe we could represent orientations as a class that could be added/subtracted, then provide a method that does both. For example:

```
image.orient(image.exif_orientation + Orientation.flip_horizontal() + Orientation.rotate(90))
```

The ``Orientation`` class would work out the final dimentions and how map the pixels across.

2) More advanced operations (such as rotation, cropping and resizing) could all be represented using affine matrices, multiplied together and run in a single pass too

Both Pillow and Wand have operations which take an affine matrix and transform the image using that matrix. We can represent transformation, rotation, scale and skew (crop being a combo of transformation and scale) using matrices and combine them, so we can perform the EXIF orientation, custom rotations, cropping and the final resizing together in a single pass.


Ideally, Wagtail would always work from the source image and apply both the changes made by the user in the admin and the developer in the template in one go. The advantages of doing it this way are:

 - If multiple rotations were applied to the image there would be less blurring of the final result
 - It shouldn't be much slower to do it this way as multiple resize/transform operations can be merged and applied together

One disadvantage is we can't merge resize/transform operations if there is a filter/effect in between (unless it's an effect that doesn't matter at what point it is applied, such as grayscale).

mrchrisadams:
Hey @kaedroho thanks for the extra comments on the image editing - you okay to make that a separate issue and address in a separate PR?

There's quite a lot going on there, and as I'll be at Wagtail space next week, it might be useful to be able to sketch it out in more detail ideally in person with someone more familiar with Willow than I am right now.

kaedroho:
> Hey @kaedroho thanks for the extra comments on the image editing - you okay to make that a separate issue and address in a separate PR?

Sure, I wasn't expecting it to be implemented in this PR. Sorry I didn't clarify that!

willow/plugins/pillow.py
mrchrisadams: Ah, _now_ I see where you're coming from - thanks for clearing that up.

I've pushed an update and a few tests for the cases you described 👍

PR #56, opened 2017-05-26
Reshaping Feature Array

trumpet2012:
From OpenCV 1 to OpenCV 2/3 there is an additional array wrapping each of the feature points. So in OpenCV 1 two feature points of [20, 40] and [30,50] would have been returned as 
```py
[
     [20,40], 
     [30, 50]
]
``` 

With the switch to numpy they were being returned as 
```py
[
     [
         [20,40]
     ], 
     [
         [30, 50]
     ]
]
``` 

Added tests to verify the values of the features/facial points to catch this in the future.


kaedroho:
Looks good, thanks again!

Issue #60, opened 2017-09-22
Tests failing with IOError: cannot write mode RGBA as JPEG


mans0954:
```
python2.7 -m unittest discover -v
...
test_save_as_jpeg (tests.test_pillow.TestPillowOperations) ... ERROR
test_save_as_jpeg_optimised (tests.test_pillow.TestPillowOperations) ... ERROR
test_save_as_jpeg_progressive (tests.test_pillow.TestPillowOperations) ... ERROR
test_save_as_png (tests.test_pillow.TestPillowOperations) ... ok
test_save_as_png_optimised (tests.test_pillow.TestPillowOperations) ... ok
test_save_transparent_gif (tests.test_pillow.TestPillowOperations) ... ok
test_transparent_gif (tests.test_pillow.TestPillowOperations) ... ok

======================================================================
ERROR: test_save_as_jpeg (tests.test_pillow.TestPillowOperations)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_pillow.py", line 31, in test_save_as_jpeg
    return_value = self.image.save_as_jpeg(output)
  File "willow/plugins/pillow.py", line 74, in save_as_jpeg
    image.save(f, 'JPEG', quality=quality, **kwargs)
  File "/usr/lib/python2.7/dist-packages/PIL/Image.py", line 1893, in save
    save_handler(self, fp, filename)
  File "/usr/lib/python2.7/dist-packages/PIL/JpegImagePlugin.py", line 604, in _save
    raise IOError("cannot write mode %s as JPEG" % im.mode)
IOError: cannot write mode RGBA as JPEG

======================================================================
ERROR: test_save_as_jpeg_optimised (tests.test_pillow.TestPillowOperations)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_pillow.py", line 39, in test_save_as_jpeg_optimised
    unoptimised = self.image.save_as_jpeg(io.BytesIO())
  File "willow/plugins/pillow.py", line 74, in save_as_jpeg
    image.save(f, 'JPEG', quality=quality, **kwargs)
  File "/usr/lib/python2.7/dist-packages/PIL/Image.py", line 1893, in save
    save_handler(self, fp, filename)
  File "/usr/lib/python2.7/dist-packages/PIL/JpegImagePlugin.py", line 604, in _save
    raise IOError("cannot write mode %s as JPEG" % im.mode)
IOError: cannot write mode RGBA as JPEG

======================================================================
[...]

PR #61, opened 2017-10-05
Implement set_background_color_rgb operation

kaedroho:
Ref: #60

This allows the user to replace the alpha channel of an image with a solid colour allowing an images with transparency to be saved JPEG without crashing (Pillow) or screwing up the image (Wand).

### DIFF
--- a/docs/guide/operations.rst
+++ b/docs/guide/operations.rst
@@ -70,12 +70,14 @@ background colour using the :meth:`~Image.set_background_color_rgb` method.
 It takes the background color as a three element tuple of integers between
 0 - 255 (representing the red, green and blue channels respectively).
 
+It returns a new :class:`~Image` object containing the background color and
+the alpha channel removed. The original image is not modified.
+
 .. code-block:: python
 
     # Sets background color to white
     i = i.set_background_color_rgb((255, 255, 255))
 
-    # This operation will remove the alpha channel from the image
     isinstance(i, Image)
     i.has_alpha() == False
 
--- a/docs/guide/operations.rst
+++ b/docs/guide/operations.rst
@@ -61,6 +61,24 @@ original image is not modified.
     isinstance(i, Image)
     i.get_size() == (200, 200)
 
+Setting a background colour
+---------------------------
+
+If the image has transparency, you can replace the transparency with a solid
+background colour using the :meth:`~Image.set_background_color_rgb` method.
+
+It takes the background color as a three element tuple of integers between
+0 - 255 (representing the red, green and blue channels respectively).
+
+.. code-block:: python
+
+    # Sets background color to white
+    i = i.set_background_color_rgb((255, 255, 255))
+
+    # This operation will remove the alpha channel from the image
+    isinstance(i, Image)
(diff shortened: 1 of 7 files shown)


willow/plugins/pillow.py
```diff
@@ -58,6 +58,24 @@ def crop(self, rect):
...
+
+        # Convert non-RGB colour formats to RGB
+        # As we only allow the background color to be passed in as RGB, we
+        # convert the format of the original image to match.
+        image = self.image.convert('RGBA')
+
+        # Generate a new image with background colour and draw existing image on top of it
+        # The new image must temporarily be RGBA in order for alpha_composite to work
+        new_image = _PIL_Image().new('RGBA', self.image.size, (color[0], color[1], color[2], 255))
+        new_image.alpha_composite(image)
```
kaedroho: Just found out this function is only available in recent versions of Pillow so we probably should have a fall back.

tomdyson:
@nealtodd any chance you could have a look at this sometime?

docs/guide/operations.rst
```diff
@@ -61,6 +61,24 @@ original image is not modified.
...
 
+Setting a background colour
+---------------------------
+
+If the image has transparency, you can replace the transparency with a solid
+background colour using the :meth:`~Image.set_background_color_rgb` method.
+
+It takes the background color as a three element tuple of integers between
+0 - 255 (representing the red, green and blue channels respectively).
+
```
nealtodd: For consistency with other method docs maybe add:

```It returns a new :class:`~Image` object containing the background color. The
 original image is not modified.```

File: docs/guide/operations.rst [hunk 3]
```diff
@@ -61,6 +61,24 @@ original image is not modified.
...
+.. code-block:: python
+
+    # Sets background color to white
+    i = i.set_background_color_rgb((255, 255, 255))
+
+    # This operation will remove the alpha channel from the image
```
nealtodd: Not sure why this is here in the context of this particular method doc?

willow/plugins/pillow.py
```diff
@@ -58,6 +58,30 @@ def crop(self, rect):
         return PillowImage(self.image.crop(rect))
 
     @Image.operation
+    def set_background_color_rgb(self, color):
+        if not self.has_alpha():
+            # Don't change image that doesn't have an alpha channel
+            return self
+
```
nealtodd: Maybe raise a descriptive exception here if `color` isn't a three-tuple (or technically an array with at least three elements unless being explicit about the type) to avoid an uncaught exception at L73

willow/plugins/wand.py
```diff
@@ -61,6 +67,23 @@ def crop(self, rect):
         return clone
 
     @Image.operation
+    def set_background_color_rgb(self, color):
+        if not self.has_alpha():
+            # Don't change image that doesn't have an alpha channel
+            return self
+
```
nealtodd: Ditto on the exception raising point in pillow method.

kaedroho:
Thanks for the review @nealtodd! I've implemented all of your suggestions now.

PR #62, opened 2018-01-17
fix save_as_jpeg failing on rgba images

temaput:
as it states here https://github.com/python-pillow/Pillow/issues/2609 we now need to explicitly convert rgba to rgb before converting to jpg

### DIFF
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -87,7 +87,7 @@ class PillowImage(Image):
 
     @Image.operation
     def save_as_jpeg(self, f, quality=85, optimize=False, progressive=False):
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks for looking into this! But you can't just convert an RGBA to RGB though as this tends to screw up images that have transparency.

Instead, I think we should use the ``set_background_color_rgb`` operation to convert the image and set the background colour to black, which I think is a reasonable default behaviour.

dopry:
@temaput would you have time to address the changes @kaedroho  requested? @kaedroho is there an underlying bug here that should be addressed? Maybe this should be moved to an issue if there is a demonstrable bug with saving to discuss a solution?

PR #63, opened 2018-02-20
Keep ICC profile and EXIF data when saving as JPEG or PNG

stefanistrate:
Pillow offers the possibility to save an ICC profile and EXIF data with an image, and Willow should keep them if already available. The ICC profile is important to preserve proper colours, while EXIF data could contain copyright information that needs to be kept.


kaedroho:
This looks good, thanks! Could you add a test?

zerolab:
This looks like it supersedes #21

stefanistrate:
I've added some tests. After checking the official Pillow documentation one more time, I realized it doesn't provide a way to save EXIF info in PNG files (http://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#png), so the latest version of my pull request adds ICC & EXIF support to JPEG, and only ICC to PNG.

trumpet2012:
@stefanistrate 	Thank you for writing this PR! We need the ability to persist the ICC profile to the renditions.

While testing we did notice however that issues can arise from persisting the EXIF data and ICC profile. See issue https://github.com/python-pillow/Pillow/issues/148

To replicate it use the following test image:
wine_taste_pix

```py
from willow.plugins.pillow import PillowImage
from willow.image import JPEGImageFile


test_file_path = ''
with open(test_file_path, 'rb') as img_file:
    image = PillowImage.open(JPEGImageFile(img_file))
    image.resize((200, 159)).save_as_jpeg('/tmp/output3.jpeg', quality=85, optimize=True, progressive=True)
```

Output:
```
Suspension not allowed here
  File "/Users/tholliday/.pyenv/versions/3.5.4/lib/python3.5/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<input>", line 3, in <module>
  File "/Users/tholliday/.local/share/virtualenvs/anvil-w_oh6ly0/src/willow/willow/plugins/pillow.py", line 107, in save_as_jpeg
    image.save(f, 'JPEG', quality=quality, **kwargs)
  File "/Users/tholliday/.local/share/virtualenvs/anvil-w_oh6ly0/lib/python3.5/site-packages/PIL/Image.py", line 1950, in save
    save_handler(self, fp, filename)
  File "/Users/tholliday/.local/share/virtualenvs/anvil-w_oh6ly0/lib/python3.5/site-packages/PIL/JpegImagePlugin.py", line 762, in _save
    ImageFile._save(im, fp, [("jpeg", (0, 0)+im.size, 0, rawmode)], bufsize)
  File "/Users/tholliday/.local/share/virtualenvs/anvil-w_oh6ly0/lib/python3.5/site-packages/PIL/ImageFile.py", line 515, in _save
    raise IOError("encoder error %d when writing image file" % s)
OSError: encoder error -2 when writing image file
```

The above code will raise an `OSError: encoder error -2 when writing image file` from PIL. Reading the linked issue, it appears that this is due to the buffer size being exceeded by the exif data. One user commented that Nikon ignored the spec for exif data size, which could explain why our test image is breaking.

It only seems to do it for certain resize sizes. For example resizing the image to `(199, 158)` will work.

Our approach is going to be to try and pass the EXIF data to Pillow and handle the exception if it blows up.

trumpet2012:
We also noticed a couple other things to consider for this PR. 

One is that ICC Profile can also cause an `OSError: encoder error -2 when writing image file` when attempting to save JPEGs. Attached is an image with an ICC profile that can cause the error.
manteca1_6-24-10

Another thing to note is that keeping the EXIF rotation data on the images can cause issues with the renditions on Wagtail. Wagtail calls `auto_orient` to rotate the image, but by keeping the EXIF rotation information on the image, some browsers like Safari will then display the image rotated again based on the EXIF.

stefanistrate:
After some years I decided to investigate this further and incorporate some of the suggested changes. I've updated this PR with the changes below. Hopefully it will be accepted and merged this time.

*  @trumpet2012 was right about the first image crashing Pillow and Willow. It still does with the most recent version of Pillow (https://github.com/python-pillow/Pillow/issues/148#issuecomment-578787435), so I integrated his code into my PR.
* I wasn't able to reproduce the crash with the ICC profile mentioned in the second image from @trumpet2012. Maybe that was happening with an old version of Pillow? I see a change in 2017 in Pillow that was supposed to address that properly: https://github.com/python-pillow/Pillow/commit/ebaa509d66049b5c21e6f786ad08a317abf9a14a#diff-3e1df49350929c75888e12b51a918bfb
* I used the recent `exif_transpose()` function from Pillow to make sure that after calling `auto_orient()`, EXIF data doesn't contain the `Orientation` tag any more.
* One check on this PR is failing because it's on python3.4, while Pillow requires python>=3.5: https://pypi.org/project/Pillow/ I think Willow should give up on python3.4 as well.

ztasre:
Is this PR dead?

BrendanBerkley:
What's the status of this?

I've been troubleshooting a problem: An art gallery client noticed that the colors lost vibrance on the website. I poked around Willow a bit, and noticed that the `optimize` kwarg didn't seem to have any impact. But in all of my exported images, the color profile has been removed.

I might investigate further, but I'm guessing this PR would fix my problem.

stefanistrate:
Hi @BrendanBerkley!

I don't have any hopes that my change will be merged into Willow, but for my personal project I'm using the altered version without problems. If you want to test whether this fixes your issue, try adding this to your `requirements.txt`:

```
git+git://github.com/stefanistrate/Willow@v1.4.post0#egg=Willow
```

Ștefan

zerolab:
@kaedroho the ICC profiles question keeps popping up, most recently being this week in the Wagtail Slack. Aside from the one Python 3.4 test failure, is there anything to consider here from your POV?

kaedroho:
@zerolab Thanks for looking at this!

Only consideration that comes to mind from me is that converting to classes like ``RGBImageBuffer`` will drop the profile and it's technically possible that these could be converted back into images and saved. I don't think we have to deal with this for this PR though as it's probably a little niche.

stefanistrate:
> @stefanistrate do have time to update this PR. I think there is maintainer interest in getting your work merged, but the failing builds were a blocker.

I don't have any free time these months to work on it. I'm happy to let anyone build on my code and submit it quicker if there's increased interest.

PR #69, opened 2018-10-30
[WIP] Image optimisation operations

zerolab:
@kaedroho this is a first pass at using image optimisation binaries.

Opening fairly early to get initial feedback.

On my To-Do list:

- [] Decide on the best place to trigger the optimisation
- [] Optimisation chain to choose the best match for a given image
- [] Possible async processing
- [] Tests
- [] Documentation


kaedroho:
This looks good so far, thanks!

I think it would be really nice if this could be fit in to an ``Image`` class somehow. For example:

```
class JpegOptim(Image):
    def __init__(self, file):
        self.file = file

    @classmethod
    @Image.converter_from(JpegImageFile)  # Will require an @converter_to(JpegImageFile) in Pillow/Wand
     def from_jpeg_file(cls, file):
        return cls(file)

    @Image.operation
    def optimize(self):
        # Loads, optimizes and saves the file
```

This has the advantage of being able to bypass Pillow/Wand if the image to be optimised is already saved as a JPEG file. But the example above suffers from some issues such as not being able to select output format or filename, and leaves choosing an optimiser up to Willow's routing (which might not be adequate for this use case). I'm happy for this to be done differently if it can't be fit into Image classes, so just take this as some food for thought.

zerolab:
That is exactly the type of feedback I was looking for. Thanks Karl.
I think it can be possible to piggyback on https://github.com/wagtail/Willow/blob/master/willow/image.py#L89

Will let it simmer for a day or two and come back it fresh

zerolab:
Note to self: [cwebp](https://developers.google.com/speed/webp/docs/cwebp) and [oxipng](https://github.com/shssoichiro/oxipng)

willow/plugins/pillow.py
```diff
@@ -150,6 +153,28 @@ def set_background_color_rgb(self, color):
 
         return PillowImage(new_image.convert('RGB'))
 
+    @Image.operation
+    def optimize(self, image_file, format, quality=85, progressive=False):
+        for optimizer in self.get_optimizers_for_format(format):
+            optimizer.process(image_file, quality=quality, progressive=progressive)
+
+    def save_and_optimize(self, f, image, format, **kwargs):
```
ababic, line 161: Suggestion: Rename this method to `save()` and make `optimize` an kwarg with a default value of `True`

willow/plugins/pillow.py
```diff
@@ -158,13 +183,15 @@ def save_as_jpeg(self, f, quality=85, optimize=False, progressive=False):
...
+        kwargs = {
+            "quality": quality
+        }
         if optimize:
             kwargs['optimize'] = True
         if progressive:
             kwargs['progressive'] = True
 
-        image.save(f, 'JPEG', quality=quality, **kwargs)
+        f = self.save_and_optimize(f, image, "JPEG", **kwargs)
```
ababic, line 194: Nitpick: We could retain the simplicity of the original link by removing `f = ` here, since `f` is written to directly by `save_and_optimize()`

willow/plugins/pillow.py
ababic: Rationale: We are always saving, but not always optimising. Also, it might be handy to retain a way to save without optimising even when optimisers are registered (for example, to compare results with/without optimisation)

ababic: Since this is an instance method and `image` is almost always `self.image`, would it be worth ditching `image` from args here and looking for an override in kwargs instead? Line below could then be:

`image = kwargs.get("image") or self.image`

willow/plugins/pillow.py
```diff
@@ -150,6 +153,28 @@ def set_background_color_rgb(self, color):
 
         return PillowImage(new_image.convert('RGB'))
 
+    @Image.operation
+    def optimize(self, image_file, format, quality=85, progressive=False):
+        for optimizer in self.get_optimizers_for_format(format):
+            optimizer.process(image_file, quality=quality, progressive=progressive)
+
+    def save_and_optimize(self, f, image, format, **kwargs):
+        optimizers = self.get_optimizers_for_format(format)
```
ababic, line 162: This should be updated to query the registry now, right?

ababic (commented):
Nice work so far @zerolab. Really love how this is shaping up. The registry is particularly lovely :)

willow/image.py
```diff
@@ -124,7 +130,60 @@ def save(self, image_format, output):
...
+        If the passed image file is a SpooledTemporaryFile or just bytes, we are converting it to a
+        NamedTemporaryFile to guarantee we can access the file so the optimizers to work on it.
+        If we get a string, we assume it's a path to a file, and will tyr use it directly.
+        """
+        optimizers = registry.get_optimizers_for_format(image_format)
+        if not optimizers:
+            return
+
+        has_named_file = False
+        if isinstance(image_file, SpooledTemporaryFile):
```
zerolab, line 148: The logic for this method is: create a NamedTemporaryFile if needed, optimize in place, and replace the passed `image_file`. If `image_file` is a string, we assume a path and try to optimize that directly

note: this is probably the most controversial bit.
Latest Wagtail passes on a `SpooledTemporaryFile` as of https://github.com/wagtail/wagtail/pull/10284.

Given willow is used in non-Wagtail contexts, I wanted to support: `BytesIO` (which is what Wagtail passed prior to https://github.com/wagtail/wagtail/pull/10284), `NamedTemporaryFile`, + string and bytes

The reason for string is that we want to support file paths (https://github.com/wagtail/Willow/issues/108). I am unsure about bytes, though.

willow/optimizers/base.py
```diff
@@ -0,0 +1,53 @@
+import logging
+import subprocess
+from typing import ClassVar, List
+
+logger = logging.getLogger("willow")
+
+
+class OptimizerBase:
+    binary: ClassVar[str] = ""
```
zerolab, line 9: note: I debated calling this `binary_name` or even `library`

docs/concepts.rst
```diff
@@ -65,3 +65,17 @@ library. The default ones are "pillow", "wand" and "opencv".
...
 allows Willow to generate a useful error message if an operation is requested
 that only exists in a plugin without an underlying library.
+
+Optimizers
+----------
+
+
+:ref:`Optimizers <custom-optimizers>` are classes that wrap image optimization libraries such as pngquant or jpegtran
+and are classes that wrap image optimization libraries such as pngquant or jpegtran and
+apply to certain image types. They are used to optimize images after they have been saved.
```
ababic, line 75:
```suggestion
:ref:`Optimizers <custom-optimizers>` are classes that wrap image optimization libraries such as pngquant or jpegtran, and apply to certain image types. They are used to optimize images after they have been saved.
```

willow/plugins/pillow.py
```diff
@@ -166,34 +166,40 @@ def set_background_color_rgb(self, color):
...
-    def save_as_jpeg(self, f, quality=85, optimize=False, progressive=False):
+    def save_as_jpeg(
+        self, f, quality=85, optimize=False, progressive=False, apply_optimizers=True
+    ):
         if self.image.mode in ["1", "P"]:
             image = self.image.convert("RGB")
         else:
             image = self.image
 
         # Pillow only checks presence of optimize kwarg, not its value
```
ababic, line 177: This comment feels a touch misplaced now... I wonder if it's outstayed it's welcome?

willow/registry.py
```diff
@@ -34,6 +38,7 @@ def __init__(self):
         self._registered_operations = defaultdict(dict)
         self._registered_converters = {}
         self._registered_converter_costs = {}
+        self._registered_optimizers: Set["OptimizerBase"] = set()
```
ababic, line 41: Do we not care about preserving the order? I suspect this could make for some weird inconsistencies between environments.

ababic (commented):
This is looking very good indeed, @zerolab - fabulous work! I've added a few things for consideration, but nothing blocking :)

willow/optimizers/base.py
zerolab: sold!

willow/plugins/pillow.py
```diff
@@ -166,34 +166,40 @@ def set_background_color_rgb(self, color):
         return PillowImage(new_image.convert("RGB"))
 
     @Image.operation
-    def save_as_jpeg(self, f, quality=85, optimize=False, progressive=False):
+    def save_as_jpeg(
+        self, f, quality=85, optimize=False, progressive=False, apply_optimizers=True
+    ):
```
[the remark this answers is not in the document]
zerolab: oh, I struggled with this one. a docblock is a nice suggestion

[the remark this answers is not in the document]
zerolab: done

willow/optimizers/cwebp.py
```diff
@@ -0,0 +1,30 @@
...
+        cls, file_path: str, progressive: bool = False
+    ) -> List[str]:
+        return [
+            "-q 80",  # compression factor. 100 produces the highest quality.
+            "-m 6",  # inspect all encoding possibilities for best file size
+            "-pass 10",  # max number of passes
+            "-mt",  # use multithreading if possible
+            file_path,
+            f"-o {file_path}",
+        ]
```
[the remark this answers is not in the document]
zerolab: I think the most we could offer is the quality setting, but I'd rather we do this once people get noisy about it :D incremental tweaks and so on

willow/registry.py
```diff
@@ -34,6 +38,7 @@ def __init__(self):
         self._registered_operations = defaultdict(dict)
         self._registered_converters = {}
         self._registered_converter_costs = {}
+        self._registered_optimizers: List["OptimizerBase"] = set()
```
laymonage, line 41: Drive-by comment: the value is still `set()`

```suggestion
        self._registered_optimizers: List["OptimizerBase"] = []
```

zerolab: 🤦🏼 I thought I updated that 🏃‍♂️

willow/registry.py
zerolab: fixed in 66678bb

docs/guide/extend.rst
```diff
@@ -159,3 +159,42 @@ It can now be registered using the :meth:`Registry.register_plugin` method:
     import newpillow
 
     registry.register_plugin(newpillow)
+
+
```
ababic, line 163:
```suggestion
```

docs/guide/optimize.rst
```diff
@@ -0,0 +1,24 @@
...
+
+They can be enabled by setting the ``WILLOW_OPTIMIZERS`` environment variable to ``true``. To enable a specific
+subset of optimizers, set the ``WILLOW_OPTIMIZERS`` environment variable to a comma-separated list of their
+library names. For example, to enable only ``jpegoptim`` and ``optipng``:
+
+.. code-block:: ini
+
+    WILLOW_OPTIMIZERS=jpegoptim,optipng
+
+
```
ababic, line 17:
```suggestion

```

tests/test_image.py
```diff
@@ -208,6 +212,75 @@ def test_save_as_foo(self):
...
+        self.image.optimize(b"outfile", "jpeg")
+        mock_process.assert_called_with("tempfile")
+        mock_unlink.assert_called_with("tempfile")
+
+    @mock.patch("willow.image.NamedTemporaryFile")
+    @mock.patch("willow.image.os.unlink")
+    @mock.patch("builtins.open", mock.mock_open(read_data=b"test"))
+    def test_optimize_with_spooled_temporary_file(
+        self, mock_unlink, mock_named_temporary_file, mock_process
+    ):
```
ababic, line 254: The argument order here appears not to match the order of the patches.

zerolab: That is correct. You can think of 

```python
@a
@b
def func():
    ...
```

as `a(b(func())` so b passes its param first, then a. in our case unlink, then named temp file

https://stackoverflow.com/a/27342171

willow/image.py
```diff
@@ -122,7 +128,61 @@ def save(self, image_format, output):
...
+            # rewind and replace the image file with the optimized version
+            image_file.seek(0)
+            with open(file_path, "rb") as f:
+                image_file.write(f.read())
+
+            if hasattr(image_file, "truncate"):
+                image_file.truncate()  # bring the file size down to the actual image size
+
+        if named_file_created:
+            os.unlink(file_path)
```
ababic, line 185: A bit delayed on this one (sorry!), it's important this happen regardless of what issues might occur following temp file creation. So, I would recommend wrapping the above lines in a `try:`, and these couple in a `finally:`.

tests/test_optimizers.py
```diff
@@ -0,0 +1,178 @@
...
+        with NamedTemporaryFile(delete=False) as named_temporary_file:
+            named_temporary_file.write(self.original_image)
+            image_file = named_temporary_file.name
+
+        self.optimizer.process(image_file)
+
+        with open(image_file, "rb") as f:
+            self.assertEqual(self.optimized_size, os.fstat(f.fileno()).st_size)
+
+        os.unlink(image_file)
```
ababic, line 87: The `try/finally` approach might be a good call in these tests too

ababic (changes requested):
Superb work on the test additions, @zerolab. That feels like the last piece of the puzzle. Just a couple of small 'cleanliness' tweaks to the file handling to consider, then we can get this mother merged!! 🤟

willow/image.py
zerolab: good call. at first I wanted it to fail loudly, but then again that is probably not a good idea for auxiliary operations. done

ababic (approved):
This will do nicely. Thanks @zerolab!

PR #70, opened 2019-01-09
Add WebP support

frmdstryr:


kaedroho:
Thanks for this! Implementation looks good on first glance. Just a few comments:

 - Should we check the Pillow/Wand versions before assuming WebP support?
 - Please could you remove the version bump, this is usually done as part of the release process
 - Please add a test image and tests for Pillow/Wand (see existing tests for example)

frmdstryr:
Thanks, I removed the version change, added tests, and added an `is_format_supported` method to check for WEBP support.

Looks like imghdr doesn't work for webp on 2.7 and 3.4.

frmdstryr:
I added code to backport imghdr webp detection for python < 3.5, not sure if it should be moved to a compat module or something but it works for all versions now.

kaedroho:
This looks good, thanks!

PR #75, opened 2019-04-12
Generic operations and get_pixel_count operation

kaedroho:
This PR implements a new feature called "generic operations". This allows third party apps to register operations that only interact with other operations and not the image directly, which means the operation could be used from multiple backends.

The first generic operation is ``get_pixel_count`` which calls ``get_size`` and ``get_frame_count`` and multiplies the results together. This operation gets automatically added on to the Pillow, Wand and OpenCV backends as all these backends support ``get_size`` and ``get_frame_count``.


willow/registry.py
```diff
@@ -26,15 +33,29 @@ class UnroutableOperationError(LookupError):
...
         self._registered_converters = dict()
         self._registered_converter_costs = dict()
 
     def register_operation(self, image_class, operation_name, func):
+        if operation_name in self._registered_generic_operations:
+            return OperationNameConflict("A generic operation already exists with the name '{0}'".format(operation_name))
+
         self._registered_operations[image_class][operation_name] = func
 
+    def register_generic_operation(self, dependencies, operation_name, func):
```
zerolab: an "operation_name, func, dependencies" order makes more sense to me.

I realise you are modelling this on `register_operation` above, but the distinction is that  the image_class is in the right place as it defines where the operation takes place...

willow/registry.py
```diff
@@ -1,6 +1,13 @@
 from collections import defaultdict
 
 
+class OperationNameConflict(Exception):
+    """
+    Raised when an operation is registered that clashes with an existing generic operation's name
```
zerolab: "Raised when the registered operation clashes with an existing operation" ?

willow/registry.py
```diff
@@ -76,6 +97,15 @@ def register_plugin(self, plugin):
             self.register_converter(converter[0], converter[1], converter[2])
 
     def get_operation(self, image_class, operation_name):
+        if operation_name in self._registered_generic_operations:
+            generic_operation = self._registered_generic_operations[operation_name]
+
+            for dependency in generic_operation['dependencies']:
+                if dependency not in self._registered_operations[image_class]:
+                    raise LookupError("Generic operation dependencies not met")
```
zerolab: It would be helpful to include which dependency (if not dependencies) are not met

kaedroho: This error doesn't actually get seen by end user's it'll be swallowed up by the ``find_operation`` method

PR #76, opened 2019-04-12
Pillow animated gifs

kaedroho:
This PR adds support for animated GIFs in the Pillow backend


willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+    def get_size(self):
+        return self.frames[0].get_size()
+
+    @Image.operation
+    def get_frame_count(self):
+        return len(self.frames)
+
+    @Image.operation
+    def has_alpha(self):
+        return self.frames[0].has_alpha()
```
m1kola: Is it possible that the first frame has no alpha channel, but some other frames do?

willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+        image_file.f.seek(0)
+        image = _PIL_Image().open(image_file.f)
+
+        frame = image
+        frames = []
+        while frame:
+            frames.append(frame.copy())
+
+            try:
+                foo = image.seek(image.tell() + 1)
```
m1kola: Do we need `foo = ` bit here?

willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+
+        if image.mode != 'RGB':
+            image = image.convert('RGB')
+
+        return RGBImageBuffer(image.size, image.tobytes())
+
+    @Image.converter_to(RGBAImageBuffer, cost=200)
+    def to_buffer_rgba(self):
+        if self.has_animation():
+            pass  # TODO: Raise warning
```
m1kola: Just to clarify: are you planning to fix these todos separately or in this PR?

willow/registry.py
```diff
@@ -241,6 +271,9 @@ def find_closest_image_class(self, start, image_classes):
         for image_class in image_classes:
             path, cost = self.find_shortest_path(start, image_class)
 
+            if path is None:
+                continue
```
m1kola: Not sure that I understand why we need this `if` now and why we didn't need it before. Could you please explain briefly?

willow/plugins/pillow.py
kaedroho: Interesting question, I wouldn't think it's possible for only part of the animation to have a transparency channel but will check!

willow/plugins/pillow.py
kaedroho: Nope

willow/plugins/pillow.py
kaedroho: Ahh, forgot I had these. I'll have another look in this PR

willow/registry.py
kaedroho: It's a crash bug that's on master but somehow never been an issue until now. My memory is a bit hazy as to why this feature caused it to surface. I'll see if I can reproduce the bug in a test.

willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+            ]
+
+        params = {
+            'save_all': True,
+            'duration': image.info['duration'],
+            'append_images': [frame.image for frame in frames[1:]]
+        }
+
+        if 'transparency' in image.info:
+            params['transparency'] = image.info['transparency']
```
zerolab: I could not find any reference of variability in frames transparency, but if it turns out to be true, this is another place to consider. also why not user `has_alpha()` as in the image operation?

willow/plugins/pillow.py
zerolab: Do we care if it is an animated gif? It is common practice to get a static image when saving to png/jpeg

kaedroho: If Willow is forced to decide which frame to use when saving the image, I think this should raise a warning. I know it's common to just pick the first frame but I think it's better to require them to explicitly state this as they might be calling `.save_as_png` without considering what to do with animated GIFs.

It also never makes sense to attempt to do face detection on an animated image. So if this happens there must be a missing check and a warning should help surface this issue.

We probably should have an operation available on all image classes to allow selecting individual frames in the image. This way, users can always explicitly choose the first frame if they want to remove animation (and this would have no effect on non-animated images).

Something like:

```python
myimage.frame(0).save_as_png()
```

zerolab: Choosing a frame sounds like a good idea. Should probably be a separate PR to keep this focussed.

danihodovic:
What's the status of this PR?

tomdyson:
@danihodovic it's awaiting review

willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+    @Image.operation
+    def save_as_gif(self, f):
+        image = self.frames[0].image
+        frames = self.frames
+
+        # All gif files use either the L or P mode but we sometimes convert them
+        # to RGB/RGBA to improve the quality of resizing. We must make sure that
+        # they are converted back before saving.
+        if image.mode not in ['L', 'P']:
+            frames = [
```
zerolab: for an RGBA image, e.g. https://upload.wikimedia.org/wikipedia/commons/thumb/4/47/%2BFinger-und-N%C3%A4gel-einer-Studentin-in-der-Mensa---Bild-001.gif/1280px-%2BFinger-und-N%C3%A4gel-einer-Studentin-in-der-Mensa---Bild-001.gif this fails with `'PillowImage' object has no attribute 'convert'` because the frames are `PillowImage` instances, rather than images.

Something like
```python
            converted_frames = []
            for frame in frames:
                frame.image = frame.image.convert('P', palette=_PIL_Image().ADAPTIVE)
                converted_frames.append(frame)

            frames = converted_frames
```

should do. This'll need a test

zerolab: Worth noting that I got "unrecognized raw mode" with Pillow 4.0 when trying to edit the uploaded image. Upgraded to  5.4.1 and it worked. I believe it is something related to https://github.com/python-pillow/Pillow/pull/2374

zerolab: The above test image consumes ~35M memory to generate the rendition shown on its edit page

gasman:
Code looks good to me. One thing I'm not clear on, though: given that `PillowAnimatedImage` can presumably do everything `PillowImage` can do and more, is there any reason for the "choose the best backend" logic in registry.py to ever prefer `PillowImage` - and if not, do we really need both classes to exist?

kaedroho:
> Code looks good to me. One thing I'm not clear on, though: given that PillowAnimatedImage can presumably do everything PillowImage can do and more, is there any reason for the "choose the best backend" logic in registry.py to ever prefer PillowImage - and if not, do we really need both classes to exist?

Good point! I've merged the two classes together now

willow/plugins/pillow.py
kaedroho: I've changed this to now check all the frames for alpha, if one or more has alpha then this operation would return `True`.

willow/plugins/pillow.py
kaedroho: Removed. Thanks for spotting!

willow/plugins/pillow.py
kaedroho: This should be solved now that `frames` is a list of `PIL.Image` objects.

I'll just add the test now.

willow/plugins/pillow.py
```diff
@@ -246,4 +254,144 @@ def to_buffer_rgba(self):
...
+        image.save(f, 'GIF', **params)
+
+        return GIFImageFile(f)
+
+    @Image.operation
+    def save_as_webp(self, f):
+        if self.has_animation():
+            pass  # TODO: Raise warning
+
+        return self.frames[0].save_as_png(f, optimize=optimize)
```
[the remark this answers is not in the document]
kaedroho: Ooh, this is weird. It also shouldn't be calling `save_as_png`...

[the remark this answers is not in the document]
kaedroho: .. but this doesn't exist anymore since `PillowAnimatedImage` has been merged into `PillowImage`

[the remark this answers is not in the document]
zerolab: That was on the old `save_as_png` method on `PillowAnimatedImage` - https://github.com/wagtail/Willow/pull/76/commits/f56e31746e50486e0ff8ec0e0a1e34a1e9bc90b5#diff-9bf9c7554d2d4846168b94a2dbc89951L309 - and is no longer relevant since it got merged into `PillowImage`

willow/plugins/pillow.py
kaedroho: Only information I could find on this is in the [GIF spec](https://www.w3.org/Graphics/GIF/spec-gif89a.txt)

Section 23 says that the fields which define transparency are assigned per graphics block, so it's possible that these settings can change per-frame but even on different areas of a single frame as well.

To check if Pillow supports multiple GIF palettes, I tried opening an image with multiple palettes and saving it again.

Before:
SmallFullColourGIF

After:
pillow-saved

I think the answer to that is clear! And looking at the Pillow source code, looks like only one palette gets used in the `_save` function so it's not possible for it to have different palettes for each block: https://github.com/python-pillow/Pillow/blob/master/src/PIL/GifImagePlugin.py#L507

(Image source https://en.wikipedia.org/wiki/GIF#/media/File:SmallFullColourGIF.gif)

kaedroho: To demonstrate a tool that can actually do this image blocking/pallette generation automatically, I generated a new GIF using the following website: https://ezgif.com/

ezgif com-gif-maker

This clearly has lower quality than the original, but also it's 6x larger! (315KB up from 54KB)

Trying out other formats, this APNG is 79KB and looks much better:

ezgif com-apng-maker

I can't show them here, but a high quality WEBP comes out at 115KB and a 70% quality WEBP (which looks roughly the same as the GIF above) is 40KB

All modern browsers except IE11 supports at least one of these formats ([apng](https://caniuse.com/#feat=apng ), [webp](https://caniuse.com/#feat=webp)). So I think it would be worth adding support for these into Willow.

willow/registry.py
kaedroho: This only crashes when there is a both an unreachable class and a reachable one for the same operation (and even then it's dependent on ordering) which is why the `TestFindOperation.test_find_operation_unreachable_from_a` test didn't pick it up. I've added a new test for this now.

kaedroho:
My plan for this is to implement both animated GIF and animated WEBP support in Willow, then add logic into Wagtail so that it always transcodes animated GIFs into animated WEBP files. This fixes issues with having to save a GIF that contains more than 256 colours (which will be inevitable if we've resized the image). See discussion in https://github.com/wagtail/Willow/pull/76#discussion_r331426263

coredumperror:
I just tried out this PR on my shop's Wagtail codebase, because the Wand-based solution causes our servers to die when someone uploads a sufficiently large or complex animated GIF.

Using this code, I was able to get one of the offending GIFs to generate renditions without crashing, but they're... broken. 

Here's the original (appologies for the gigantic GIF):
funnel_pluviation

And here's the Image Chooser rendition:
funnel_pluviation max-165x165

Maybe I've done something wrong? Is there anything I can tweak to try to make Pillow generate animated GIFs in a different way, which hopefully won't break them?

kaedroho:
Hey @coredumperror 

I'm getting similar results with a different image [here](https://github.com/wagtail/Willow/pull/76#discussion_r331426263)

I think this is just a Pillow limitation. It must be difficult saving GIF images like this which have a limit of 256 colours. I think the best solution would be for us to add support for animated webp/apng so images can be saved in those formats instead.

coredumperror:
> I think the best solution would be for us to add support for animated webp/apng so images can be saved in those formats instead.

Yes, please! Animated GIF is soooo 1990s. Hell, mosts sites that let you share GIFs just convert to mp4 video with no audio track. I wonder if that would be possible?

kaedroho:
> Hell, mosts sites that let you share GIFs just convert to mp4 video with no audio track. I wonder if that would be possible?

Hadn't thought of that before! I think we could view MP4 w/o audio as just another animated image format, so could support that if Pillow/wand does (but can always add an ffmpeg plugin if not!)

gasman:
Relevant blog post: https://paulbakaus.com/2020/01/17/gifs-must-die/
Historically Wagtail has deliberately shied away from dealing with video because doing it well (handling arbitrary format conversions, extracting metadata, generating thumbnails, dealing with large file uploads...) would be as big a job as the content management side... but given that the demand for animated GIFs isn't going away, and the border between animation and video is getting ever more blurred, I think it might be time to rethink that.

coredumperror:
> I think it might be time to rethink that.

I am exceedingly glad to hear this. :)

jamesray:
@zerolab possibly a duplicate of #80 , is there any traction on this?

coredumperror:
We're still very interested in Wagtail properly supporting animated whatevers (GIF, webp, mp4, etc.).

zerolab:
Incidentally @thenewguy makes the case for animated GIFs to be treated as "media" (i.e. no resizing and other Willlow filters) - https://github.com/torchbox/wagtailmedia/issues/127
Might that be the way to go?

kaedroho:
I like that idea! In many cases, GIFs would be both smaller and better quality if we don't try to process them

PR #87, opened 2020-11-23
Add travis jobs for ppc64le. Install libmagickwand-dev for ppc64le

gururajrkatti:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: python
+arch:
+    - amd64
+    - ppc64le
 
 # Test matrix
 python:
@@ -12,7 +15,18 @@ matrix:
     - python: 3.7
       dist: xenial
       sudo: true
+    - python: 3.7
+      dist: xenial
+      arch: ppc64le
+      sudo: true
 
+before_install:
+  - >
+    if [[ "$TRAVIS_CPU_ARCH" == ppc64le ]]; then
+      sudo apt-get update;
+      sudo apt-get -y install libmagickwand-dev;
+    fi 
+    
 # Package installation
 install:
   - pip install Pillow Wand


gururajrkatti:
Hi, just wanted to check if you got a chance to look into this.

tomdyson:
Thanks @gururajrkatti!

gasman:
Given that the Willow library itself is pure Python code, is it really necessary to test it on multiple architectures? Any build failures would be the responsibility of the native code dependencies (Pillow, Libmagick, Python itself...), so testing them here is presumably just replicating the tests that exist on those projects.

gasman:
...Also, Travis CI seems to be rapidly killing off their support for OSS projects (https://twitter.com/james_hilliard/status/1336081776691843072), so this is probably a bad time to add any new infrastructure that relies on them.

gururajrkatti:
Hi, I understand your point. I agree that if it is pure python code, it ideally should work on all architectures. However, in our past experience we have run into architecture specific issues even with pure python packages with their tests failing in some cases , mostly because of an indirect dependency on native code or libraries via other python dependencies. by enabling the travis support here, we are attempting to catch those issues early, hope that makes some sense? Adding @gerrith3 for more details regarding travis support for OSS projects.

gerrith3:
@gasman sorry - there is some confusion over at travis as a result of this shift to a credit based system but everything we've heard and seen from them directly, they are not backing away from their committment to open source and making travis available to open sourcec ocmmuniites.  We are working with them to get this sorted soon.

PR #92, opened 2022-02-21
Gracefully handle out-of-bounds crops

gasman:
When passed a crop rectangle that extends outside an image's boundaries, Pillow gracefully returns an image of the requested size (padding with background / transparent pixels as required), but Wand throws a value error. If the calling app is sloppy about dimensions (as is apparently the case in https://github.com/wagtail/wagtail/issues/7766), this will result in errors that only surface when Wand is in use (i.e. for animated gifs).

This PR gives the two backends a consistent "middle ground" behaviour that hopefully makes intuitive sense: the crop rectangle is clamped to the image boundary so that if it overlaps, the final image is smaller than requested. Totally out of bounds or nonsensical parameters (e.g. left > right) will raise a BadImageOperationError.

Fixes https://github.com/wagtail/wagtail/issues/7766, although we'll also want to fix the rounding error that leads to the out-of-bounds condition in the first place.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -140,12 +140,18 @@ Here's a full list of operations provided by Willow out of the box:
     (Pillow/Wand only)
 
     Cuts out the specified region of the image. The region must be a sequence of
-    four integers (top, left, right, bottom):
+    four integers (left, top, right, bottom):
 
     .. code-block:: python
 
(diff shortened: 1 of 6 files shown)


kaedroho (approved):
This looks good to me, thanks!

Issue #90, opened 2021-10-02
Support for HEIC/heif images


hpoul:
Any plans on supporting heic images?
I guess the first problem will be that `imghdr` does not support heif images, and it [doesn't look like there are any plans on supporting it](https://bugs.python.org/issue38170).

My workaround right now is to use [pyheif-pillow-opener](https://github.com/ciotto/pyheif-pillow-opener) to convert the images to jpg during upload. Are there any better options? 🤔️

bigcat88:
maybe a `pillow-heif` will be better.  i'm pretty sure it will be. =)

bigcat88:
@hpoul do you still need this?
I can made a PR to allow accepting HEIF files to Pillow, OpenCV and probably Wand too.

gasman:
Implemented in #105

PR #96, opened 2022-04-03
replace deprecated imghdr with filetype

hpoul:
Since [imghdr is deprecated](https://peps.python.org/pep-0594/#imghdr) replace it with `filetype` package.
This would allow adding support for more image file types like [heic](https://github.com/wagtail/Willow/issues/90).


kaedroho (approved):
Thanks for implementing this! Looks good to me

PR #97, opened 2022-06-11
Fix Pillow 10 deprecation warnings

tomkins:
Noticed these appearing on a more recent project with Pillow>=9.1.0.

I've done a bunch of try/except to keep compatability with older versions of Pillow, as >=9.1.0 supports Python 3.7 as a minimum.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
         "filetype>=1.0.10,!=1.1.0",
     ],
     extras_require={"testing": [
-        "Pillow>=6.0.0,<10.0.0",
+        "Pillow>=9.1.0,<11.0.0",
         "Wand>=0.6,<1.0",
         "mock>=3.0,<4.0",
     ]},
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -64,7 +64,8 @@ class PillowImage(Image):
         else:
             image = self.image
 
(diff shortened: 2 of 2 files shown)


tomkins:
As the project has been updated to only support Python 3.7+, let me know if you'd like me to rework it to only support Pillow >=9.1.0 and reduce the number of try/excepts.

tomkins:
As older versions were deprecated in #98 - I've rebased and updated this PR to focus on Pillow >=9.1.0.

https://github.com/wagtail/Willow/commit/7cd344950675351da57107948705c33b5c853feb is still around with try/except if supporting older versions of Pillow is preferred.

zerolab:
The changes as they are LGTM ([ref](https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#deprecations)).

@kaedroho any reason we'd want to support Pillow < 9.1.0?

dopry:
I vote to drop the old PIL version support. Using the latest version should give us a more consistent feature set to work with when using PIL. Moving to 9.2 would be ideal since some [underlying deps of 9.1 are EOL](https://pillow.readthedocs.io/en/stable/releasenotes/9.2.0.html)

kaedroho:
@zerolab We discussed this today and think we should drop support for older versions of Pillow. I can't remember a single time their API has been broken in a major release and there's often scary looking security issues so probably a good idea to push people to upgrade

gasman (approved):

PR #102, opened 2022-10-19
feat: add mime type  to ImageFile

dopry:
Add mime_type to ImageFile to facilitate
generating Content-Type headers for images

While working on https://github.com/wagtail/wagtail/pull/8974 @jams2 noted that Pillow provided a content_type that was lost in the switch to Willow. I implemented a mapping of format_name to mime_type to resolve the issues, but it would be better if we could get the mime_type from the ImageFile instance returned by willow.open

### DIFF
--- a/willow/image.py
+++ b/willow/image.py
@@ -1,6 +1,6 @@
-import filetype
 import warnings
 
+import filetype
 from filetype.types import image as image_types
 
 from .registry import registry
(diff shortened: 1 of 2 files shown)


dopry:
@kaedroho @jams2 @zerolab I'd love a review on this. It's pretty small, but it is an interface addition.

willow/image.py
```diff
@@ -135,6 +135,7 @@ def has_animation(self):
 
 class ImageFile(Image):
     format_name = None
+    mime_type = None
```
zerolab, line 138: question: wouldn't `application/octet-stream` make sense for an unknown type

[RFC 2046](http://www.rfc-editor.org/rfc/rfc2046.txt) states in section 4.5.1:

> The "octet-stream" subtype is used to indicate that a body contains arbitrary binary data.

dopry: I think mime_type should be required in the ImageFile implementation. Not assigning it should generally create an error.  Maybe I should reimplement this as a getter that throws and exception by default?

dopry: - now raises a NotImplementedError if mime_type or format_name are missing from an ImageFile implementation. 
- added test coverage

zerolab (approved):

PR #103, opened 2022-10-20
chore: update PIL 10 deprecations

dopry:


zerolab:
Thank you for this @dopry. It is a duplicate of #97, though.
Could you please review/weigh in on #97?

PR #105, opened 2022-10-27
Adding support for HEIF for Pillow. #90

### LINKED ISSUE #90: Support for HEIC/heif images

bigcat88:
Resolves #90


gasman:
Merged in 716f0613296eea85a61a22d8c8828a5cb5a94c03 (with docs).

PR #117, opened 2023-06-26
Switch to using flit for packaging

zerolab:
Builds on #116

PR #118, opened 2023-06-27
Use RTD config v2, update Sphinx configuration and use the Wagtail sphinx theme

zerolab:
Builds on #117 and

* adds an RTD configuration file (ref: https://blog.readthedocs.com/migrate-configuration-v2/)
* tidies up the sphinx configuration
* switches to using https://github.com/wagtail/sphinx-wagtail-theme


docs/conf.py
```diff
@@ -1,274 +1,66 @@
...
-# Add any paths that contain templates here, relative to this directory.
 templates_path = ["_templates"]
+exclude_patterns = ["_build", ".DS_Store"]
+
+html_theme = "sphinx_wagtail_theme"
+html_theme_path = [sphinx_wagtail_theme.get_html_theme_path()]
+html_theme_options = {
+    "project_name": "Willow",
+    # "logo": "",
+    "github_url": "https://github.com/wagtail/Willow/tree/main/docs",
```
Stormheg, line 40: **Suggestion:** add a trailing slash to generate the correct urls. Without it, you'll get urls like `https://github.com/wagtail/Willow/blob/main/docsinstallation.rst`
```suggestion
    "github_url": "https://github.com/wagtail/Willow/tree/main/docs/",
```

zerolab: Going to https://github.com/wagtail/Willow/tree/main/docs doesn't redirect to https://github.com/wagtail/Willow/tree/main/docs/ so both are valid?

docs/conf.py
```diff
@@ -1,274 +1,66 @@
...
-
-# If true, keep warnings as "system message" paragraphs in the built documents.
-# keep_warnings = False
-
+html_static_path = ["_static"]
 
-# splhinxcontrib.spelling settings
+pygments_style = None  # covered by sphinx_wagtail_theme
 
 spelling_lang = "en_GB"
```
Stormheg, line 49: **Suggestion:** per Wagtail core team discussions, we are adopting American English for our docs. I think this should apply to Willow too.

ref: https://github.com/wagtail/wagtail/issues/10608

zerolab: Done 86f3cc8

docs/conf.py
Stormheg: When I build the docs locally and click the 'edit on github' button, it takes me to a 404 page. This happens because incorrect urls are generated. Adding a trailing slash to the url like suggested resolves the issue.

Stormheg: In case it is not clear: Sphinx uses the `github_url` setting to generate links for the 'edit on github' button.

zerolab: ah.. good to know. sorted

Stormheg (approved):
Looking mighty fine - thanks!

PR #128, opened 2023-08-04
Fix `NUMBER_PATTERN` regex for parsing SVG viewboxes

### LINKED ISSUE #127: SvgViewBoxParseError at /cms/images/
### Issue Summary

Wagtail images breaks completely when it cannot parse an svg image correctly. Example given.
Error:
`SvgViewBoxParseError at /cms/images/
Unable to parse viewBox value '-5.000000011410315e-06 0.0 750.00001 525.000007'`

Notice the -5.000000011410315e-06 value which is likely the culprit. However, the svg validates as valid.

### Steps to Reproduce

Upload the attached svg image, then in Wagtail go to "Images". Or try to add one to a page using the chooser.

### Technical details

-   Python version: 3.11
-   Django version: 4.2
-   Wagtail version: 5.0
-   Browser version: irrelevant
Home_hero
Auswahl_847

jams2:
Fixes #127 

The `NUMBER_PATTERN` in the SVG viewbox regular expression was missing the optional `+` or `-` in the final [integer](https://www.w3.org/TR/SVG11/types.html#DataTypeInteger) production.

The definition for `number`, as used in SVG attributes, can be found here:
- https://www.w3.org/TR/SVG11/types.html#DataTypeNumber
- https://www.w3.org/TR/2013/WD-SVG2-20130409/types.html#DataTypeNumber


tests/test_svg_image.py
```diff
@@ -417,6 +417,32 @@ def test_view_box_re(self):
             (".1,0,0,0", ViewBox(0.1, 0, 0, 0)),
             ("+.1,0,0,0", ViewBox(0.1, 0, 0, 0)),
             ("-.1,0,0,0", ViewBox(-0.1, 0, 0, 0)),
+            # https://github.com/wagtail/Willow/issues/127
+            (
+                "-5.000000011410315e-06 0.0 750.00001 525.000007",
+                ViewBox(-5.000000011410315e-06, 0.0, 750.00001, 525.000007),
```
zerolab, line 423: can you add a test for `-5.000000011410315e06` which should be valid as far as I can tell from https://www.w3.org/TR/SVG11/types.html#DataTypeNumber and https://www.w3.org/TR/SVG11/types.html#DataTypeInteger ?

Other than that, LGTM

jams2: Good point, added a few more tests.

PR #129, opened 2023-09-05
Ensure SVG files are given a mime type

RealOrangeOne:
`SvgImageFile` currently doesn't conform to the `ImageFile` interface, as it doesn't define `mime_type`.

It might be nice to define `ImageFile` as a proper ABC, but that's a conversation for another day. There are a few stream types too which will need to be considered


zerolab (approved):
LGTM

PR #130, opened 2023-10-09
Add optional dependencies for Pillow/Wand

zerolab:
so one can install `Willow[Pillow]` and get the correct Pillow dependencies


Stormheg (approved):
LGTM

PR #132, opened 2023-10-25
test_image.py: replace wrong unicdoe character

stephanlachnit:
The character U+0441 "с" was changed to the ASCII character U+0063 "c".

### DIFF
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -182,7 +182,7 @@ class TestImageFormats(unittest.TestCase):
         self.assertIsInstance(image, HeicImageFile)
         self.assertEqual(width, 320)
         self.assertEqual(height, 241)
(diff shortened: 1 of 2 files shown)


zerolab:
@stephanlachnit that's a good catch. There are a couple of other places where this needs updating

```
~ ❯❯❯ ag "с" .
tests/test_image.py
185:        self.assertEqual(image.mime_type, "image/heiс")
220:            self.assertEqual(image.mime_type, "image/heiс")

build/lib/willow/image.py
267:        return "image/heiс"

willow/image.py
327:        return "image/heiс"
```

annoyingly the GitHub interface doesn't differentiate between them :/

stephanlachnit:
> There are a couple of other places where this needs updating

Good catch, I updated the PR.

PR #135, opened 2023-10-26
test_detect_faces: sort array for determinism

### LINKED ISSUE #134: test_opencv.TestOpenCVOperations.test_detect_faces is flaky
I had a case where the test failed one time with the following error:
```
FAIL: test_detect_faces (tests.test_opencv.TestOpenCVOperations.test_detect_faces)
----------------------------------------------------------------------
  File "/tmp/autopkgtest.BPuIbH/autopkgtest_tmp/tests/test_opencv.py", line 74, in test_detect_faces
    assert_allclose(faces, self.expected_faces, atol=2)
  File "/usr/lib/python3/dist-packages/numpy/testing/_private/utils.py", line 1592, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/numpy/testing/_private/utils.py", line 862, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=2

Mismatched elements: 8 / 8 (100%)
Max absolute difference: 182
Max relative difference: 1.98901099
 x: array([[ 90, 164, 188, 262],
       [272,  88, 365, 181]], dtype=int32)
 y: array([[272,  89, 364, 181],
       [ 91, 165, 187, 261]])
```

As can be seen, this is essentially the same array just ordered differently. I think in the context of detecting faces that should be okay? I don't know if there is a why to test this though, maybe one can sort the outer array.

Most of the time, the test does not seem to fail.

stephanlachnit:
Closes #134

This only sorts the order of the tuples, no the values in the tuples itself.

### DIFF
--- a/tests/test_opencv.py
+++ b/tests/test_opencv.py
@@ -71,4 +71,5 @@ class TestOpenCVOperations(unittest.TestCase):
         self.assertIsInstance(faces, list)
         # There are two faces in the image
         self.assertEqual(len(faces), 2)
-        assert_allclose(faces, self.expected_faces, atol=2)
+        # Sort out list as there is no preferred order of the faces
+        assert_allclose(sorted(faces), sorted(self.expected_faces), atol=2)


zerolab (approved):
I was in two minds at first, but after some more digging, this makes sense.

PR #136, opened 2023-11-02
Fix color management by keeping ICC color profiles and EXIF data in addition

### LINKED ISSUE #114: Faulty color profile conversion
### Issue Summary

Photo files with an AdobeRGB color profile are not correctly converted. The result looks less saturated. The resulting images (JPG and WebP format) use a color profile, named "RGB". If I upload the source files with an "sRGB" profile, the results match.

Is there a setting to tell Pillow how to convert these color spaces/profiles? "Relative colorimetric" should be the way to go, I think. I would love to see a setting to configure this properly. "sRGB" could always be the color space for the converted images. But I would love to see more perceptually true results from other color spaces as well. Many users do not know about these technical details at all.

Many other websites seem to handle this better, like you can see on the right image in the screenshot.
The difference seems to be more emphasized in Safari on Mac OS. But also in Firefox on Linux I was seeing the difference.

### Technical details

-   Python version: 3.10.6
-   Django version: 3.2.13
-   Wagtail version: 4.2

Screenshot 2023-06-21 at 15 23 58

Left: Original JPG with AdobeRGB color profile
Middle: Wagtail rendition
Right: Rendition of contemporaryartdaily.com for comparison

### LINKED ISSUE #137: AVIF image saved with incorrect EXIF orientation metadata after calling `auto_orient`
When calling the `auto_orient()` method on a JPEG image with EXIF orientation metadata, the resulting image has the orientation applied directly to the image data. This works as expected.

But if the image is then saved using `save_as_avif()`, the (now incorrect) EXIF orientation metadata is encoded into the resulting AVIF file as well. This incorrectly tells viewers to rotate the image, resulting in a end user seeing a rotated image.

## Steps to reproduce

1. Create a new Django project: `django-admin startproject willow_avif_issue`
2. Download and unzip [image.jpeg.zip](https://github.com/wagtail/Willow/files/13404319/image.jpeg.zip) in the root of the django project created in the previous step. The extracted `image.jpeg` has EXIF orientation metadata. Here's a screenshot of the orientation metadata from macOS Preview:
image

3. Install Willow with AVIF support in your Django project: `pip install willow[heif]`
4. Create a new management command (`touch willow_avif_issue/management/commands/test_willow.py`)  with the following contents. Willow will throw an error if it is not being called from inside a configured Django project.

```python
# willow_avif_issue/management/commands/test_willow.py
from django.core.management.base import BaseCommand
from willow.image import Image


class Command(BaseCommand):
    help = 'Test willow'

    def handle(self, *args, **options):
        with open('./image.jpeg', 'rb') as f:
            img = Image.open(f)

            img = img.auto_orient()
            img.save_as_avif('image.avif')
```

5. Run the management command: `python manage.py test_willow`

6. The resulting `image.avif` created in the project root should have the EXIF orientation directly applied to the image data. This can be verified by viewing the width and height. The `image.avif` also retains the same EXIF orientation metadata. This tells viewers to rotate the image (again), resulting in a rotated image:

image

## Expected outcome

The EXIF orientation metadata is no longer correct and should probably be removed.

## Notes

This specifically affects saving as AVIF. When saving as JPEG, it appears EXIF metadata is being (mostly) stripped. Other formats don't support orientation metadata at all I believe.

### Technical details

-  Python version: 3.11.6
-  `pip freeze` output
```
asgiref==3.7.2
defusedxml==0.7.1
Django==4.2.7
filetype==1.2.0
Pillow==10.1.0
pillow-heif==0.13.1
sqlparse==0.4.4
Willow==1.6.2
```

andre-fuchs:
Adds methods to keep the ICC color profile and EXIF data if supported for JPEG, PNG, WebP and AVIF file formats. Extends the unit tests accordingly. All tests passed.

Closes #114
Closes #137


zerolab:
@andre-fuchs thank you very much for this.

We use pre-commit with black/ruff, and is one of the testing dependencies - https://github.com/wagtail/Willow/blob/main/pyproject.toml#L44

Can you run `pre-commit install` then `pre-commit run --all-files` to lint?

andre-fuchs:
Thanks @zerolab. Just did learn about pre-commit and reformatted the code.

zerolab:
Thank you @andre-fuchs. Will aim to review and test ASAP

Stormheg:
The changes to `auto_orient` will also end up fixing #137.

Two birds with one stone!

Stormheg (approved):
LGTM!

willow/plugins/pillow.py
```diff
@@ -166,6 +168,26 @@ def set_background_color_rgb(self, color):
 
         return PillowImage(new_image.convert("RGB"))
 
+    @Image.operation
```
zerolab, line 171: The second point is we try to keep parity where possible between `PillowImage` and `WandImage`, so the changes would need reflected in `willow.plugins.wand.WandImage`

Reference: https://docs.wand-py.org/en/latest/guide/exif.html - (`image.profiles["EXIF"]`, `images.profiles["ICC"]`)

willow/plugins/pillow.py
```diff
@@ -287,63 +319,39 @@ def save_as_heic(
...
-                            Image.Transpose.FLIP_LEFT_RIGHT,
-                        ),
-                        6: (Image.Transpose.ROTATE_270,),
-                        7: (Image.Transpose.ROTATE_90, Image.Transpose.FLIP_LEFT_RIGHT),
-                        8: (Image.Transpose.ROTATE_90,),
-                    }
-
-                    for transpose in ORIENTATION_TO_TRANSPOSE[orientation]:
-                        image = image.transpose(transpose)
+        image = ImageOps.exif_transpose(self.image)
```
zerolab, line 354: This and related tests are fantastic 👏🏼

zerolab (changes requested):
@andre-fuchs sorry for the delay. Lots of fantastic work here! Added a few notes on improving the API a bit. 

Do let me know if you have the capacity to address these

willow/plugins/pillow.py
```diff
@@ -166,6 +168,26 @@ def set_background_color_rgb(self, color):
 
         return PillowImage(new_image.convert("RGB"))
 
+    @Image.operation
+    def add_icc_profile(self, image, kwargs):
```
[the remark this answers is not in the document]
andre-fuchs: Please review my last commit which resolves this issue.

willow/plugins/pillow.py
andre-fuchs: Thanks for pointing this out. I would have missed the Wand implementation completely.

ICC color profiles and EXIF data are not saved to all formats by Wand. Pillow manages to get ICC color profiles embedded in PNG and AVIF files, for example, but Wand does not in its current implementation. Am I missing something here? 

**This works:**
JPEG with ICC color profiles and EXIF data
WebP with ICC color profiles

zerolab: Will have to dig into this later today.
Given Wand is a wrapper around ImageMagick, we are likely limited by that.

andre-fuchs: In the `save_as_jpeg` and `save_as_webp` methods of the `WandImage` class, I use the `.clone()` method, which wasn’t used before here. I think this is necessary, as Wand’s/ImageMagick’s operations are destructive (in contrast to Pillow/PIL) and I want to preserve the original image here.

willow/plugins/wand.py
```diff
@@ -164,8 +170,18 @@ def save_as_jpeg(
         :param apply_optimizers: controls whether to run any configured optimizer libraries
         :return: JPEGImageFile
         """
-        with self.image.convert("pjpeg" if progressive else "jpeg") as converted:
+        with self.image.clone() as converted:
```
zerolab: @andre-fuchs I changed this back to use `convert` as https://docs.wand-py.org/en/latest/wand/image.html#wand.image.Image.convert as that returns a new image, rather than do anything on the original

zerolab (approved):

andre-fuchs:
Thanks @zerolab for tidying up!

zerolab:
Thank you for helping us take this over the finish line, @andre-fuchs!
It has been way too long coming

andre-fuchs:
Cannot wait to use it. Many of my clients were complaining about the color conversion in Wagtail. This should solve it.

laymonage:
@andre-fuchs You can install version 1.6.3 instead: https://github.com/wagtail/Willow/releases/tag/v1.6.3. Wagtail's upper version boundary on Willow likely won't be updated until Wagtail 6.0 unless #11272 is backported to the `stable/5.2.x` branch.

PR #138, opened 2023-11-19
Clean up `auto_orient` logic

### LINKED ISSUE #137: AVIF image saved with incorrect EXIF orientation metadata after calling `auto_orient`

Stormheg:
Move away from bespoke logic and rely on `ImageOps.exif_transpose()`. This method works much better than our own as it also supports reading XMP metadata instead of just EXIF.

In addition, it also removes the orientation metadata from the transposed image.

This method was added in Pillow 6.0.0.
See: https://github.com/python-pillow/Pillow/blob/28c173f8d4767c7f6dd22dc840117fe641f4d3ee/docs/releasenotes/6.0.0.rst#added-imageopsexif_transpose

This fixes #137 in the process by removing stale orientation metadata.

### DIFF
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -29,6 +29,12 @@ def _PIL_Image():
     return PIL.Image
 
 
+def _PIL_ImageOps():
+    import PIL.ImageOps
+
+    return PIL.ImageOps
+
+
 class PillowImage(Image):
     def __init__(self, image):
         self.image = image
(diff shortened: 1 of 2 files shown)


Stormheg:
Removing code feels so good 😛
image

willow/plugins/pillow.py
```diff
@@ -29,6 +29,12 @@ def _PIL_Image():
     return PIL.Image
 
 
+def _PIL_ImageOps():
+    import PIL.ImageOps
+
+    return PIL.ImageOps
+
```
Stormheg, line 36: I'm following the same pattern as the code above. Not sure if it is needed though, I assume the indirection is to defer loading large modules? Not sure if that applies to `PIL.ImageOps`

Stormheg:
As it turns out, a very similar change is proposed in #136. Closing this in favour of that PR!

PR #141, opened 2024-01-10
Add support for ICO images

RealOrangeOne:
Especially useful when dynamically creating favicons.

`wagtail.ico` is taken from wagtail.org.


zerolab (approved):
I was in two minds about this. However ICO files are still widely used and Willow is used outside just Wagtail so.. OK, since the footprint is tiny

Issue #139, opened 2023-12-21
Inaccurate colors / broken WEBP/AVIF/HEIC images with specific image files


Stormheg:
A client uploaded some images to their Wagtail site and reported these images failed to load on the site or displayed with inaccurate colors.

The images in question all meet the following conditions:

- The image is in CMYK color mode (which I believe is rare for images on the web, but very common for images meant for print)
- The image has an embedded ICC color profile

**Willow fails to handle images that meet the above conditions correctly.**

I took me quite a while to figure out why this is happening, but here is what I believe to be happening:

The AVIF, HEIC and WEBP Pillow encoders _don't_ support CMYK. Before the image can be saved in those formats, it has to be converted to RGB first. This conversion fails to take into account the color profile.

<img width="1167" alt="Screenshot 2023-12-13 at 2 28 56 PM" src="https://github.com/wagtail/Willow/assets/13856515/c6d6baf6-062e-4958-b432-baed735330a5">

Here is a screenshot comparison of the color issue. On the left is the original image, on the right is the WEBP rendition created using Pillow 1.6.3. It is subtle, but if you pay close attention you can see the WEBP version has somewhat brighter colors. This is most noticable around the headlamp and the lens flare near the hand.

For AVIF and HEIC specifically, the resulting file fails to display at all. Something funky has happened corrupting the image. No error is thrown however.

I've been experimenting with my fork of Willow and have found that forcing a conversion to sRGB color space (and implicitly to RGB) using Pillows' `ImageCms.profileToProfile` before saving works very well. The result looks much, much, closer to the original image (if rendering_intent=perceptual is used). The aforementioned issue with AVIF and HEIC is also fixed by this.

##  `transform_colorspace_to_srgb` operation

As part of my fix, I've been working on a general `transform_colorspace_to_srgb` operation for Willow. The idea behind this is that most devices cannot display colors outside of sRGB and perform some sort of transformation to sRGB already when confronted with wide-gamut images (like DCI-P3 or Adobe RGB). This operation allows performing this transformation upfront which in turn allows the profile to be removed, resulting in a smaller image file. 

I think it makes sense to make this functionality available in Willow. Maybe Wagtail could use this operation in the future to create even smaller image renditions.

The downside of course is that this is lossy operation, devices that _can_ display color spaces like DCI-P3 get an image with less color but that seems like a reasonable trade-off for the majority which cannot. I guess there could always be a setting in Wagtail to disable this behaviour or make it a different kind of rendition filter or something.

[...]

PR #142, opened 2024-01-14
Add `transform_colorspace_to_srgb` operation and use it to fix inaccurate colors when saving specific image files

Stormheg:
Please see https://github.com/wagtail/Willow/issues/139 for a detailed description of what this is trying to accomplish. Took me longer than expected to prepare a PR, writing tests turned out to be quite time consuming!

I've expanded the automatic conversion discussed in #139 to include `png` files too as they are also affected by the color issue described in #139. Technically, GIF files are also affected but I've not bothered to fix them because there isn't much to gain when the colors are butchered to fit in a palette of 256 colors.

The automatic conversion during save of `avif`, `heic`, `webp` and `png` images _only_ happens when we encounter an image that is affected by the issue (CMYK mode, with embedded ICC profile). For all other images, the new code path is not taken and there should be no impact.

I'd appreciate reviews from anyone! (not necessarily Willow maintainers only).

I've done my best to explain how and why in the docs and comments but I'm no expert on the subject of color spaces and gamuts and such. Please let me know if there are any inaccuracies or ambiguities.


tests/test_pillow.py
```diff
@@ -124,6 +125,72 @@ def test_set_background_color_rgb_color_argument_check(self):
...
+        # deviations from the expected result without being too sensitive to minor
+        # differences between systems.
+        stat = ImageStat.Stat(image_srgb.image)
+        expected_sum2 = [1286116001.0, 1222827100.0, 1144012271.0]
+
+        for actual, expected in zip(stat.sum2, expected_sum2):
+            self.assertAlmostEqual(
+                actual,
+                expected,
+                delta=1000,
```
Stormheg, line 185: I've assumed that the results on my system might not be the same as on other systems. Hence the the `assertAlmostEqual`.

I haven't actually run this test on any other systems so not sure if my assumption is actually true.

Stormheg:
I've added a new image for the test suite: `dog_and_lake_cmyk_with_icc_profile.jpg`. I've modified this image specifically to be affected. The image has been heavily downsized to make its size as small as possible but the embedded ICC profile still takes up a lot of space. It is a little over 1MB in size.

I took the image a couple years ago during a hike with my dog. Willow has my permission to use the downsized image file for its test suite.

Here is a full-sized image for your viewing pleasure. Try opening and saving it as `webp` or `avif` with an unpatched version of Willow and this patch to see the difference.
[zaar-lake_iso_coatedv2.jpg.zip](https://github.com/wagtail/Willow/files/13932094/zaar-lake_iso_coatedv2.jpg.zip)

image

_screenshot comparing the original JPEG on the left with an unpatched WEBP version._

willow/plugins/pillow.py
```diff
@@ -227,19 +278,32 @@ def save_as_png(self, f, optimize: bool = False, apply_optimizers: bool = True):
...
+                pillow_image = self.transform_colorspace_to_srgb()
+                image = pillow_image.image
+                kwargs["icc_profile"] = pillow_image.get_icc_profile()
+            else:
+                kwargs["icc_profile"] = icc_profile
+
         if self.image.mode == "CMYK":
             image = self.image.convert("RGB")
         else:
             image = self.image
```
zerolab, line 302: This now does double duty. I believe this now need to be

```suggestion
        elif self.image.mode == "CMYK":
            # we don't have an ICC profile, and the image is in CMYK, so convert to RGB
            image = self.image.convert("RGB")
```

tests/test_pillow.py
zerolab: Let's try to be more strict and see what the CI says

zerolab (commented):
Nice one @Stormheg 
Left a couple of notes

andre-fuchs:
@Stormheg Very cool! I suggest to add a system wide setting to convert all images to sRGB. Two setting variables would desirable, I think. A boolean type to turn this conversion on or off. And another one for the rendering intend (could be both an integer or char type). This is mainly an update of Wagtail, but I guess that Willow also would need some preparation for that. Does this make sense?

Stormheg:
@andre-fuchs lets discuss automatically converting to sRGB in a separate issue 👍 

Since most images and devices consuming the web are only capable of displaying sRGB this should be mostly fine; the space saving would be worth the cost in most cases I think. I could imagine that a photography site might not want to perform such conversions however. Saturated sRGB colors look a bit bland compared to what is possible with P3 and Adobe RGB.

Maybe there could be a middle ground where we provide functionality to detect what color gamut the original image is in? And if it is in a color other than sRGB, Wagtail could output something akin to the snippet below. This will only load the wide gamut with original ICC profile for devices that support displaying that color gamut.

```html
<picture>
  <source media="(color-gamut: p3)" srcset="photo-wide.jpg">
  <img src="photo-srgb.jpg">
</picture>
```

Somewhat niche functionality maybe, but with the number of devices that are capable of recording and displaying wide-gamut images increasing (in particular recent iPhones are capable of this) it would be good future proofing for Wagtail to be aware of wide gamut images.

Anyway, let's discuss separately from this PR 😄

willow/plugins/pillow.py
Stormheg: Little oversight of mine, thanks!

tests/test_pillow.py
Stormheg: CI is green with no delta, I'll make this much simpler.

zerolab (approved):

zerolab:
Nice one @Stormheg. Let's discuss the suggestion from @andre-fuchs separately

PR #148, opened 2024-06-19
Better handle types when running optimisers

### LINKED ISSUE #147: Dynamic image preview fails when WILLOW_OPTIMIZERS is set
### Issue Summary

Currently when you've got an optimizer enabled in `WILLOW_OPTIMIZERS`, and use the Image URL Generator, it'll come up with a 500:

```
Internal Server Error: /admin/images/26/preview/original/
  ...
  File "/Users/tomkins/Python/wagtail/wagtail/images/views/images.py", line 295, in preview
    image = Filter(spec=filter_spec).run(image, response)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tomkins/Python/wagtail/wagtail/images/models.py", line 1031, in run
    return willow.save_as_png(output, optimize=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tomkins/Python/willow/willow/plugins/pillow.py", line 313, in save_as_png
    self.optimize(f, "png")
  File "/Users/tomkins/Python/willow/willow/image.py", line 178, in optimize
    optimizer.process(file_path)
                      ^^^^^^^^^
UnboundLocalError: cannot access local variable 'file_path' where it is not associated with a value
```

Slightly obscure (not sure how many people actually the combination of Willow optimizers and the dynamic image serve view!).

### Steps to Reproduce

- Install optipng if you don't have it already (`brew install optipng`)
- Add `ServeView` into your URL config (see https://docs.wagtail.org/en/v6.1.2/advanced_topics/images/image_serve_view.html#setup)
- Edit your settings file, add: `WILLOW_OPTIMIZERS = "optipng"`
- Upload a PNG into Wagtail images (eg. https://github.com/wagtail/wagtail/blob/d44808ac532a96c60b16377fbb25d29d7eae704c/docs/logo.png)
- Go to the edit image view, click on URL Generator on the right hand side
- Page loads, but the image fails to appear and generates the above 500

-   I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

-   Python version: 3.12 (any)
-   Django version: 4.2 (any)
-   Wagtail version: Current main
-   Browser version: n/a

### Working on this

Pull request coming shortly!

RealOrangeOne:
Fixes #147 

The main change here is adding an explicit type error, rather than having a variable not be defined.

Secondly, the type handling itself gets simpler and slightly more optimised.

### DIFF
--- a/willow/image.py
+++ b/willow/image.py
@@ -1,6 +1,7 @@
 import os
 import re
 from io import BytesIO
+from shutil import copyfileobj
 from tempfile import NamedTemporaryFile, SpooledTemporaryFile
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


willow/image.py
```diff
@@ -147,42 +148,44 @@ def optimize(self, image_file, image_format: str):
...
-                        named_file.write(file.read())
-                    file_path = named_file.name
-                named_file_created = True
-            elif isinstance(image_file, BytesIO):
-                with NamedTemporaryFile(delete=False) as named_file:
-                    named_file.write(image_file.getvalue())
+                    named_file_created = True
+
+                    image_file.seek(0)
+                    copyfileobj(image_file, named_file)
```
zerolab, line 156: 👌🏼 
oh yes, this makes this whole block so much simpler

willow/image.py
```diff
@@ -147,42 +148,44 @@ def optimize(self, image_file, image_format: str):
...
+
             elif isinstance(image_file, bytes):
                 with NamedTemporaryFile(delete=False) as named_file:
                     named_file.write(image_file)
                     file_path = named_file.name
                     named_file_created = True
 
+            else:
+                raise TypeError(
+                    f"Cannot optimise {type(image_file)}. Must be a readable object, or a path to a file"
```
zerolab, line 174: nitpick: 
```suggestion
                    f"Cannot optimise {type(image_file)}. It must be a readable object, or a path to a file"
```

willow/image.py
RealOrangeOne: `shutil.copyfileobj` is definitely one of those gems that makes things a lot cleaner. Sure, it's quite literally just buffered `.read` and `.write` calls - but it makes our code cleaner!

zerolab (approved):

zerolab:
Needs backporting to `stable/1.8.x`

PR #152, opened 2024-09-05
Fix `test_process_optimizes_image`

Stormheg:
On my machine, one optimized version is 56kb smaller which surpasses the allowed tolerance but not in a negative way.

This PR addresses that issue uses `assertAlmostEqual` instead because it's delta parameter is clearer than a factor.


zerolab (approved):
no complaints from me

PR #155, opened 2024-09-13
Bump actions to latest

Stormheg:
Hoping this will resolves the CI failures we are seeing regarding uploading and downloading of coverage data. Update latest Python while at it.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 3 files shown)


zerolab (approved):
LGTM.

there are a few more things we can do (like setting `COVERAGE_CORE=sysmon` when running on Python 3.12+ to speed up coverage) but they can be sorted later

Stormheg:
@zerolab sounds cool, I can do that now as I have some time available for Willow today.

PR #160, opened 2024-12-24
Fix Github Actions for Ubuntu 24.04 base image

Stormheg:
Github actions has now defaulted to using its Ubuntu 24.04 base image.
Compared to the previous base image, this is missing many pre-installed
binaries. Such as imagemagick.

Comparison:

- Ubuntu 24.04: https://github.com/actions/runner-images/blob/ubuntu24/20241215.1/images/ubuntu/Ubuntu2404-Readme.md
- Ubuntu 22.04: https://github.com/actions/runner-images/blob/ubuntu24/20241215.1/images/ubuntu/Ubuntu2204-Readme.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,8 +34,12 @@ jobs:
         python: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
-      - name: Install optimizers
+      - name: Install required system packages
         run: |
+          sudo apt-get update -y
+          # For Wand tests
+          sudo apt-get install -y imagemagick
+          # For testing optimizers
           sudo apt-get install -y jpegoptim pngquant gifsicle optipng libjpeg-progs webp
       - name: Set up Python ${{ matrix.python }}
         uses: actions/setup-python@v5
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
           python-version: ${{ env.PYTHON_LATEST }}
       - uses: pre-commit/action@v3.0.1
   test:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Stormheg:
Not sure what causes the CI to fail... Can't seem to reproduce this locally in a Ubuntu 24.04 docker container.

.github/workflows/test.yml
```diff
@@ -27,15 +27,19 @@ jobs:
...
       matrix:
         python: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
-      - name: Install optimizers
+      - name: Install required system packages
         run: |
+          sudo apt-get update -y
+          # For Wand tests
+          sudo apt-get install -y imagemagick
```
zerolab, line 41: given 22.04 has imagemagick, do we want this?

zerolab: or.. perhaps try installing `libmagick++-dev libmagickcore-dev libmagickwand-dev` in addition to imagemagick on 24.04?

Stormheg: @zerolab I tried to ascertain that the test suite does pass on on 22.04, which it does. That manual install is leftover from trying to fix the 24.04 build.

Not sure why it is failing on 24.04. I know that 24.04 does not ship with ImageMagick installed, so we do that manually. That does not help however, there is something wrong where it fails somewhat silently by returning None.

I'd like to get to the bottom of this, but don't have a lot of capacity to invest at the moment. Perhaps we should pin the workflow to 22.04 for now so the testsuite continues working and create an issue to move to 24.04

zerolab: let's do that. solve the immediate problem, then spend more time looking into 24.04

Stormheg:
Closing this in favor of stop-gap https://github.com/wagtail/Willow/pull/162

PR #163, opened 2025-02-25
Run OpenCV tests in GitHub Actions

Stormheg:
Noticed we were not actually running OpenCV tests.


zerolab (approved):
LEt's get this in as well

PR #165, opened 2025-02-25
Run tests against Python 3.14 prerelease

Stormheg:
We already have the `allow-prerelease` flag set to allow prereleases: https://github.com/wagtail/Willow/blob/d77b646848f405e7c78b704af0b84b434ddaf500/.github/workflows/test.yml#L43-L47

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
     needs: lint
     strategy:
       matrix:
-        python: ["3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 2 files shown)


zerolab:
It looks like we need to install extra libs when running with 3.14 - https://pillow-heif.readthedocs.io/en/latest/installation.html#linux - until pillow_heif provides wheels

Stormheg:
@zerolab support was added to pillow_heif three weeks ago: https://github.com/bigcat88/pillow_heif/releases/tag/v1.1.0

Pillow 11.3 already shipped with 3.14 wheels. Looks like we are all set ✅

zerolab (approved):
🤘

Issue #166, opened 2025-03-19
AVIF support in pillow-heif is deprecated


laymonage:
pillow-heif v0.22.0, released a few days ago, has deprecated AVIF support:

https://github.com/bigcat88/pillow_heif/releases/tag/v0.22.0

The package now points to https://github.com/fdintino/pillow-avif-plugin instead.

On the other hand, there has been movement in the PR to add AVIF support to Pillow upstream (https://github.com/python-pillow/Pillow/pull/5201). That PR might land soon, so we might want to either ignore the warning for now or set an upper bound on pillow-heif to `<0.21.0` instead of migrating to pillow-avif-plugin.

PR #167, opened 2025-04-01
Use native Pillow v11.2.0 AVIF format

Stormheg:
Fixes https://github.com/wagtail/Willow/issues/166

Pillow v11.2.0 was released today with AVIF support as a notable new feature. 

Requires raising the minimum version of Pillow. Only the latest Pillow release receives active support so requiring the latest version doesn't seem like a big issue to me. Especially now that AVIF support in pillow-heif is deprecated, as noted in https://github.com/wagtail/Willow/issues/166

### DIFF
--- a/willow/plugins/pillow.py
+++ b/willow/plugins/pillow.py
@@ -1,7 +1,7 @@
 from io import BytesIO
 
 try:
-    from pillow_heif import AvifImagePlugin, HeifImagePlugin  # noqa: F401
+    from pillow_heif import HeifImagePlugin  # noqa: F401
 except ImportError:
     pass
 
@@ -425,7 +425,12 @@ class PillowImage(Image):
     def save_as_avif(self, f, quality=80, lossless=False, apply_optimizers=True):
         kwargs = {"quality": quality}
         if lossless:
-            kwargs = {"quality": -1, "chroma": 444}
+            kwargs = {
+                # Quality of 100 implies lossless (according to libavif documentation)
+                "quality": 100,
+                # When encoding lossless images, don't use chroma subsampling (4:4:4 retains the most information)
+                "subsampling": "4:4:4",
+            }
 
         image = self.image
         icc_profile = self.get_icc_profile()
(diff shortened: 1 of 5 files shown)


willow/plugins/pillow.py
```diff
@@ -425,7 +425,12 @@ def save_as_heic(
     def save_as_avif(self, f, quality=80, lossless=False, apply_optimizers=True):
         kwargs = {"quality": quality}
         if lossless:
-            kwargs = {"quality": -1, "chroma": 444}
+            kwargs = {
+                # Quality of 100 implies lossless (according to libavif documentation)
+                "quality": 100,
+                # When encoding lossless images, don't use chroma subsampling (4:4:4 retains the most information)
+                "subsampling": "4:4:4",
```
Stormheg, line 432: This won't actually be valid when encoding a grayscale image, should be `4:0:0` instead (grayscale is Pillow `L` or `LA` mode image) - something for a follow up?

zerolab: let's do it as a follow up (as long as you have the time :D)

Stormheg: Tried writing a failing test but it turns out `save_as_avif` will convert from `LA` to `RGBA` internally - which makes this not really relevant... or at least not without same yak shaving to figure out why the conversion is happening.

Let's leave this here for now.

Stormheg:
Marking as draft for now, this needs more work (upstream)

The Pillow 11.2.0 release was retracted because the added AVIF encoders/decorders severely increased the size of prebuilt wheels, causing the project to hit its PyPI upload limit which resulted in not all platforms having a prebuilt 11.2.0 wheel available. As a result, 11.2.1 was released which no longer ships with the `libavif` library included by default. [1] Including `libavif` requires building from source, which is a hassle.

This poses an issue for this PR as we can no longer assume AVIF support is available.

The Pillow maintainers have indicated that they will revisit the issue in 11.3.0, so let's cross our fingers and hope they manage to fix the issue [2].

[1]: https://github.com/python-pillow/Pillow/blob/7a0092f2072a5deca433e2a6b752eced49e4ee16/docs/releasenotes/11.2.1.rst?plain=1#L6-L10
[2]: https://github.com/python-pillow/Pillow/pull/8876#issuecomment-2798906089

Stormheg:
Closing, superseded by #170

PR #168, opened 2025-04-22
Fix broken AVIF support when used with Pillow 11.2.1

Stormheg:
See @laymonage's comment: https://github.com/python-pillow/Pillow/issues/8886#issuecomment-2813504451
This applies the fix proposed on that discussion by Andrew Murray - thank you

Also pinning pillow_heif for now so our users don't end up pulling in a version of that package that no longer has AVIF support. We should either move to https://github.com/fdintino/pillow-avif-plugin (unsure what the future of that plugin looks like now that AVIF support has technically shipped in Pillow) or wait for Pillow 11.3.0 due for release next quarter which has the potential of shipping full AVIF support

See also #167 for more background

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,8 +34,9 @@ dependencies = [
 pillow = ["Pillow>=9.1.0,<12.0.0"]
 wand = ["Wand>=0.6,<1.0"]
 heif = [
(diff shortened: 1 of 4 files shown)


zerolab (approved):
LGTM

PR #170, opened 2025-07-03
 Use native Pillow v11.3.0 AVIF format

Stormheg:
Supersedes #167 

Pillow 11.3.0 does ship with AVIF support libraries included in the wheels at a reasonable file size - building from source is not required anymore.

This allows us to adopt this method, meaning we no longer have to rely on the deprecated support in pillow-heif. As a result that library's version can be unpinned 🎉 

IMPORTANT: as a consequence, we do have to raise the minimum version of Pillow. I believe this is fine as only the latest Pillow versions are officially supported.


zerolab (approved):

kuflierl:
I would love seeing this merged. As the maintainer for libheif for nixpkgs i would like to upgrade it to 1.20. This causes breakage for pillow-heif. Pillow heif does work if upgraded to 1.0.0 and patched (424448). But Willow seems to only support pillow 0.22 at the moment.

Stormheg:
I am planning to make on tomorrow.

Willow doesn't follow Wagtail's release schedule, so any fixes necessary (like this one) can be released out of band.

Stormheg:
@kuflierl v1.11 has now been released to PyPI: https://pypi.org/project/Willow/1.11.0/

PR #185, opened 2026-07-07
Various maintenance bits

Stormheg:
Various maintenance bits and improvements.

- Replace trove license classifier with updated `license` key syntax in pyproject.toml, per [PEP 639](https://peps.python.org/pep-0639/#add-license-expression-field)
- Introduce Zizmor to Willow (already running for Wagtail)
- Update pre-commit hooks
- Add pyproject-fmt hook

### AI usage

Some GitHub Copilot auto-completion

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -54,5 +54,3 @@ jobs:
 
       - name: 🚀 Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
-        with:
-          packages-dir: ./dist
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,11 +11,11 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.14"
           cache: "pip"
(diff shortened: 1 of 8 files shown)


.github/workflows/publish.yml
```diff
@@ -41,16 +45,14 @@ jobs:
...
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: dist
           path: ./dist
 
       - name: 🚀 Publish package distributions to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
         with:
           packages-dir: ./dist
```
zerolab: aside: the default for `packages-dir` is `dist` so this could be dropped too

pyproject.toml
```diff
@@ -1,128 +1,106 @@
...
+  "E",   # pycodestyle errors
+  "F",   # pyflakes
+  "I",   # isort
+  "T20", # flake8-print
+  "UP",  # pyupgrade
+]
+# E501: Line too long
+lint.ignore = [ "E501" ]
+
+[tool.coverage]
```
zerolab: non-blocking: I like to keep this in its own `.coveragerc` file. saves a few bytes from end package size =)

pyproject.toml
```diff
@@ -1,128 +1,106 @@
...
+run.branch = true
+run.omit = [ "tests/*", "willow/utils/deprecation.py" ]
+run.source_pkgs = [ "willow" ]
+paths.source = [ "willow" ]
+report.exclude_lines = [
```
zerolab: worth using [`exclude_also`](https://coverage.readthedocs.io/en/7.15.0/config.html#report-exclude-also) so we extend the defaults and take advantage to any future defaults https://coverage.readthedocs.io/en/7.15.0/excluding.html#default-exclusions

zerolab (approved):
LGTM, with a couple of non-blocking changes

pyproject.toml
Stormheg: Thanks for making me aware of this cool new feature in coverage.py!

zerolab (approved):

PR #186, opened 2026-07-08
Fix OpenCV 5+ compatibility

Stormheg:
### Description

In the recently released OpenCV 5.0, `cv2.CascadeClassifier` has moved to the opencv-contrib package ([relevant migrating to v5.0 section](https://github.com/opencv/opencv/wiki/OpenCV-4-to-5-migration#objdetect-haar-and-hog-moved-to-contrib)). Adjust our dependency installation code accordingly and add a message for end-users that might run into it.

### AI usage

None

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
           cache: pip
       - name: Install dependencies
         run: |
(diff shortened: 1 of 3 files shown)


Stormheg:
Replaced by #187

PR #187, opened 2026-07-08
Drop OpenCV haarcascade files

### LINKED ISSUE #65: [RFC] Make OpenCV a separately installed module?
Just want to gather opinions on whether we should move the OpenCV plugin (along with the face/feature detection operations) into a separate module. This should be easy to do as it's decoupled already.

Reasons:
 - I don't think it's very widely used
 - The majority of the size of Willow's distributions on pypi is taken by one file that's only required by OpenCV https://github.com/wagtail/Willow/blob/master/willow/data/cascades/haarcascade_frontalface_alt2.xml
 - There is an alternative available (https://github.com/torchbox/rustface-py) that's much easier to install than OpenCV. Having OpenCV built in to Willow core would make it less likely for someone to try out alternatives which could be better.

Drawbacks:
 - Breaks backwards compatibility
 - Will make face detection features less discoverable

Stormheg:
Fixes #65 

### Description

We can use the haarcascade data files that ship with OpenCV instead. It is my assumption that when face detection support with opencv was added to Willow, these files didn't ship with OpenCV or weren't stable enough to depend on. With haarcascades being 'old tech' these days there isn't much that is going to change anymore.

### Compatibility implications

None, this change is backwards compatible.

### AI usage

None


zerolab:
Worth noting that when we upgrade to opencv 5, we will want to use the new DNN-based face detector
https://github.com/opencv/opencv/wiki/OpenCV-4-to-5-migration#objdetect-haar-and-hog-moved-to-contrib

zerolab (approved):

Stormheg:
Thanks for the review Dan!

PR #188, opened 2026-07-09
Add JPEG-XL open/save support

### LINKED ISSUE #184: Add support for uploading JPEG-XL .jxl files
### Problem statement

I am creating an archival system for artists. I am strongly encouraging the use of JXL files for upload as a means of getting archival quality at low file sizes.

I have managed to get support for this format working, but it took some hoops.
1. Re-registering JXL with filetype (no idea why, as it is included, but I had to do it anyway).
2. monkey-patch get_accept_attributes for the file-picker
3. monkey-patch Image.open
4. monkey-patch PillowImage init, format_name and mime_type

This is probably a terrible way of getting there.

### Proposed solution

Enable this in core Willow, checking for the (current) dependency on pillow_jxl which would likely be many fewer lines of code in implementation.

JXL is hopefully making its way to the web (again), but for now it is useful as an upload format.

### Alternative solutions

_No response_

### Additional context

_No response_

### Working on this

Contribute to the discussion to help this move forward. Explain your use cases, discuss possible solutions, research how to adapt existing code, or compatibility concerns.

Once others agree with the proposed changes, anyone can contribute to this. View our [contributing guidelines](https://docs.wagtail.org/en/latest/contributing/index.html). Add a comment to the issue to explain what approach you are taking.

Stormheg:
Fixes #184 

Relates to https://github.com/wagtail/roadmap/issues/235 on the roadmap

### Description

Adds support for reading and writing .jxl (JPEG-XL) files through https://github.com/Isotr0py/pillow-jpegxl-plugin/.

Optimizer support is also available through `cjxl`

### Reviewer notes

I vendored in a patch to https://github.com/h2non/filetype.py since that hasn't had a release shipping in 4 years. I don't see a release happening soon and don't want this PR to be blocked waiting for filetype.py indefinitely. We can, in due time when a release is made, remove this patch.

### AI usage

VSCode editor autocomplete through GitHub Copilot


willow/plugins/pillow.py
```diff
@@ -268,6 +274,63 @@ def save_as_jpeg(
...
+                pillow_image = self.transform_colorspace_to_srgb()
+                image = pillow_image.image
+                image.info["icc_profile"] = pillow_image.get_icc_profile()
+            else:
+                image.info["icc_profile"] = icc_profile
+
+        elif image.mode == "CMYK":
+            image = image.convert("RGB")
+
+        image.save(f, "JXL", **kwargs)
```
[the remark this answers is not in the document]
Stormheg: Fair callout, but pillow_jxl actually handles this automatically for us by reading `image.info["exif"]` etc.. We don't need to explicitly pass anything like we do with the other image formats.

https://github.com/Isotr0py/pillow-jpegxl-plugin/blob/1851b06a4a48f0604d55bf08df89264ac9b643d4/pillow_jxl/JpegXLImagePlugin.py#L155-L167

willow/image.py
```diff
@@ -256,6 +256,16 @@ def mime_type(self):
...
 
 
+class JXLImageFile(ImageFile):
+    @property
+    def format_name(self):
+        return "jxl"
+
+    @property
+    def mime_type(self):
+        return "image/jxl"
```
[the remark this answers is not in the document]
Stormheg: Good robot 🤖  Fixed!

Stormheg:
@zerolab when you have availability, could you review this please? No rush. Keep calm 🧘

willow/plugins/pillow.py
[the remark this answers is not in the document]
zerolab: thought: this could be an inline comment. Something along the lines of: "Unlike other formats, pillow_jxl handles EXIF metadata automatically, so we don't need to explicitly pass it"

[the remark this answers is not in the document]
Stormheg: Addressed in a469b2467e38fcc2d69fad303ca8a8a83c9a6559

docs/guide/optimize.rst
```diff
@@ -1,10 +1,13 @@
 Optimizing images
 =================
 
-Aside from the basic optimizations that the Pillow backend provides, Willow supports using dedicated libraries to
-optimize images. Out of the box, Willow comes with optimizers for `gifsicle <https://www.lcdf.org/gifsicle/>`_,
-`jpegoptim <https://www.kokkonen.net/tjko/projects.html>`_, `optipng <https://optipng.sourceforge.net/>`_,
-`pngquant <https://pngquant.org/>`_ and `cwebp <https://developers.google.com/speed/webp/docs/cwebp>`_.
+Aside from the basic optimizations that the Pillow backend provides, Willow
+supports using dedicated libraries to optimize images. Out of the box, Willow
+comes with optimizers for `gifsicle <https://www.lcdf.org/gifsicle/>`_,
```
zerolab: just noticed this. I think 
```suggestion
comes with support for `gifsicle <https://www.lcdf.org/gifsicle/>`_,
```

would make more sense semantically.

non-blocking

docs/guide/save.rst
```diff
@@ -24,13 +25,12 @@ For example, to save an image as a PNG file:
         i.save_as_png(f)
 
 Changing the quality setting
----------------------------------
+----------------------------
```
zerolab, line 28: 👌🏼

tests/test_optimizers.py
```diff
@@ -83,9 +84,15 @@ def test_process_optimizes_image(self):
...
             with open(image_file, "rb") as f:
-                self.assertAlmostEqual(
-                    self.optimized_size, os.fstat(f.fileno()).st_size, delta=60
-                )
+                processed_image_size = os.fstat(f.fileno()).st_size
+
+                # If the image is larger than we expect based on the reference
+                # target size we have, allow a delta of 60 bytes to account for
+                # any minor differences in optimization.
+                if processed_image_size > self.optimized_size:
```
zerolab, line 92: thought: we want `assertAlmostEqual` to always run

and, if for some reason the processed image size is vastly smaller than what we anticipate, it is something to check (e.g. did the library improve so much that we need to adjust our baseline? or is there an error?)

tests/test_optimizers.py
```diff
@@ -179,3 +186,52 @@ def get_check_library_command_arguments(self):
...
+
+@unittest.skipUnless(Cjxl.check_library(), "cjxl not installed")
+class CjxlOptimizer(DefaultOptimizerTestBase, TestCase):
+    extension = "jxl"
+    optimizer = Cjxl
+
+    def test_applies_to(self):
+        self.assertTrue(Cjxl.applies_to("jxl"))
+        for ext in ("png", "jpeg", "gif", "tiff", "bmp"):
+            self.assertFalse(Cjxl.applies_to(ext))
```
zerolab: nitpick: normally for iterative test with also do `self.subTest` with an appropriate message

zerolab (changes requested):
A couple of small changes needed, otherwise LGTM!

Thank you @Stormheg

tests/test_optimizers.py
Stormheg: My mac running `cjxl v0.12.0 0.12.0 [_NEON_] {AppleClang 21.0.0.21000099}` creates jxl files that are more than the allotted 60 bytes smaller. I believe the difference was almost 300 bytes.

The version of cjxl shipped with the Ubuntu 26.04 runner is older and presumably doesn't feature the same optimizations that later versions have. The reference optimized image was created using the version that ships with Ubuntu 26.04 to set a baseline under which the CI can pass.

I don't want to increase the delta for `assertAlmostEqual` to a value that large, especially since an image becoming larger than the reference optimized version is cause for concern. If it less than the reference, I don't find that very concerning. Since the version of the optimizer used is not directly under our control, I'd rather not have to adjust our baseline as external tooling improves, keep maintenance low.

Thinking about it, we should probably assert the image is not zero bytes in size. But I don't think we should be running `assertAlmostEqual` always.

@zerolab maybe we should add a `processed_image_size < self.optimized_size` case and check the size is not more than 10% off from our reference? Instead of comparing absolute byte sizes.

tests/test_optimizers.py
Stormheg: Fixed for this new test and other similar tests in 91597e1

tests/test_optimizers.py
zerolab: Not more than 10% sounds like a plan

Stormheg: Done in 98d09f252ef7352ac262b8aa85532183e08c160f - marking this resolved 👍

Stormheg:
@zerolab let me know if you are happy for this to be merged or want more time to review 👍

I'd like to get https://github.com/wagtail/Willow/pull/190 in too before I make a new release.

engineervix:
I recently encountered an error in an automated process that involves extracting images from pdf and uploading them to the image library.

```
Failed to upload image on page 1: Cannot load jpx images

{
  'jpg': <class 'willow.image.JPEGImageFile'>,
  'png': <class 'willow.image.PNGImageFile'>,
  'gif': <class 'willow.image.GIFImageFile'>,
  'bmp': <class 'willow.image.BMPImageFile'>,
  'tif': <class 'willow.image.TIFFImageFile'>,
  'webp': <class 'willow.image.WebPImageFile'>,
  'svg': <class 'willow.image.SvgImageFile'>,
  'heic': <class 'willow.image.HeicImageFile'>,
  'avif': <class 'willow.image.AvifImageFile'>,
  'ico': <class 'willow.image.IcoImageFile'>
}
```

Which led me to this PR. 

From <https://jpeg.org/jpeg2000/> and <https://jpeg.org/jpegxl/>, it looks like `.jpx` isn't part of the JPEG XL spec, it's still part of the JPEG 2000 spec, but falling under "Part 2: Extensions". 

If that's ok, I can create a separate PR to add  `.jpx` support.

Stormheg:
@engineervix JPEG 2000 is an old format and superseded in its capabilities by modern formats like AVIF and JPEG XL. It does make sense you found images in this format inside a PDF (presumably the PDF was created during the same era as JPEG 2000 😉 )

I don't see sense in adding support for such an old format to Wagtail. In your specific case, my recommendation is to convert to a more conventional format before uploading to Wagtail.

zerolab:
@Stormheg @thibaudcolas had this saved from the weekend - https://giannirosato.com/blog/post/case-against-jxl/. Sharing without comment

PR #190, opened 2026-07-12
Revamp release workflow

Stormheg:
### Description

Revamp release workflow to harden security. 

- Pushing a git tag will trigger a wheel and source dists to be build and signed. They are uploaded to a draft GitHub release, including an appropriate extract from the release notes.
- After human review the release can be published. The publish.yml workflow then downloads the release artifacts, checks they were created from GitHub actions and publishes them.
- CHANGELOG.rst converted to markdown and put in the root of the project, with symlink so it is still build by Sphinx
- RELEASING.md documents this release process for us and future maintainers.

I verified these changes work as expected by publishing a release on my own fork: https://github.com/Stormheg/Willow/actions.

Loosely based on what I did for another package I maintain: https://github.com/stormbase/django-otp-webauthn/

Happy to discuss changes.

### AI usage

RELEASING.md was entirely AI-generated by ChatGPT based on same vague do-this-do-that instructions.


RELEASING.md
```diff
@@ -0,0 +1,165 @@
...
+
+- the release version is correct
+- the release notes are correct
+- the attached wheel and source distribution files are present
+
+If the release is incorrect:
+
+1. Delete the draft GitHub Release.
+2. Fix the issue.
+3. Push a new tag (you may replace the previous tag if necessary).
```
zerolab, line 107: quick note: we have immutable releases enabled, so this would fail

Image

Stormheg: If the draft release is removed, it would be possible to modify the tag, no? As long as there is not a release pointing to that tag.

zerolab: hmm, you're right. you are talking about drafts.
(sorry about the noise!)

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          persist-credentials: false
+
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.x"
+
+      - name: Install build dependencies
+        run: |
+          python -m pip install flit
```
Stormheg, line 35: **Thought:** we could consider pinning this to a known-good version. But then we would have to keep it updated too. Would the added security be worth it?

zerolab: The publish action also doesn't pin it. one for a follow up

.github/workflows/publish.yml
```diff
@@ -1,56 +1,60 @@
 name: Publish to PyPI
```
zerolab, line 1: I think it would be worth adding a job to push to test.pypi.org (say when we merge to `main`). This would alse be a good test of the attestations flow.

A bit further down the line, it may be worth checking out https://github.com/hynek/build-and-inspect-python-package/ to see if it simplifies things

zerolab (approved):
LGTM.

Added a note about a test.pypi.org but that can come in a separate PR

